Issue #16, opened 2016-03-14
Some labels are not displayed properly


essobedo:
I don't know if I miss something or not but when I launch `gradle run` and/or `gradle build`, some labels are not displayed properly indeed for example instead of _Edit_ I have _$[edit]_.

essobedo:
Here is a fix proposal #17

PR #17, opened 2016-03-14
#16: Resolve the variables on the fly

essobedo:
A fix proposal for https://github.com/mucommander/mucommander/issues/16:

This PR allows to resolve the variables of type _$[var-name]_ at initialization time thanks to a decorator.

### DIFF
--- a/src/main/java/com/mucommander/text/Translator.java
+++ b/src/main/java/com/mucommander/text/Translator.java
@@ -122,16 +122,16 @@ public class Translator {
         // Determines if language is one of the languages declared as available
         if(availableLanguages.contains(locale)) {
             // Language is available
-			resourceBundle= ResourceBundle.getBundle("dictionary", locale);
+            resourceBundle= ResourceBundle.getBundle("dictionary", locale);
             LOGGER.debug("Language "+locale+" is available.");
         }
         else {
             // Language is not available, fall back to default language
-			resourceBundle= ResourceBundle.getBundle("dictionary");
+            resourceBundle= ResourceBundle.getBundle("dictionary");
             LOGGER.debug("Language "+locale+" is not available, falling back to English");
         }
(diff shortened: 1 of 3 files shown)


src/main/java/com/mucommander/text/Translator.java
```diff
@@ -222,36 +220,14 @@ public static String get(String key, String... paramValues) {
...
-                                buffer.append(value, matcher.start(), matcher.end());
-                            }
-                            startIndex = matcher.end();
-                        }
-                        buffer.append(value.substring(startIndex));
-                        result = buffer.toString();
-                        cache.putIfAbsent(key, (String) result);
-                    }
-                }
+                return resourceBundle.getObject(key);
```
ahadas: so do we still need the code in lines 229-231 and not just to return whatever is stored in the map?

ahadas:
very nice contribution! probably an issue that caused by the move to properties files.. minor comment inside. sorry for the long time it took me to get to it. please address the comment inside and lets merge it

src/main/java/com/mucommander/text/Translator.java
essobedo: @ahadas Actually we still need it as it only keep key/value pairs whose value contains at least one variable as you can see [here](https://github.com/mucommander/mucommander/pull/17/files#diff-1517cfcc43a86b331edf0ab843b6b0b1R280)

essobedo: @ahadas Actually we still need it as it only keep key/value pairs whose value contains at least one variable as you can see [here](https://github.com/mucommander/mucommander/pull/17/files#diff-1517cfcc43a86b331edf0ab843b6b0b1R280)

ahadas:
Thank you for this contribution.

Issue #21, opened 2016-04-13
Locales not working on OS X 10.11.4 with Java 1.8.0_72


sebastian-lempert:
I am using OS X 10.11.4 with Java 1.8.0_72. When building muCommander (head, checked out 5 minutes ago from https://github.com/mucommander/mucommander) with

`./gradlew createDmg`

the resulting binary contains strings / locales that are not working / that are not correctly displayed. I am using German / Deutsch, the corresponding translation should be contained in the file dictionary_de.properties but other languages seem to have the same problem.

Are the files that can be found at https://github.com/mucommander/mucommander/tree/master/src/main/resources still correct? When viewing these files in a browser (Firefox 45.0.2) they are not correctly displayed too (for example see https://github.com/mucommander/mucommander/blob/master/src/main/resources/dictionary_de.properties).

Maybe it is an issue with the corresponding file encoding? If so the following information might be useful: The terminal command

`file -I dictionary_de.properties `

gives

`dictionary_de.properties: text/plain; charset=unknown-8bit`

ahadas:
Fixed

PR #22, opened 2016-04-16
Fix dictionary encoding issues in non english languages

mkortkamp:
- this fix corrects the broken encoding in non english languages (see sebastian-lempert's recent mail on the mucommander mailing list)
- the cause was that with one of the last changes before spliting the dictionary.txt the old single file was saved in non UTF-8
- all dictionaries are restored as UTF-8 from this old version and all changes since then are applied again
- Translator.java is changed to load these files as UTF-8
- gradle eclipse import hook was added to automatically set the correct encoding of these files


mkortkamp:
fixes Issue #21

src/main/resources/dictionary_ZH_CN.properties
```diff
@@ -1,8 +0,0 @@
-#pack.error_on_file = 压缩时发生错误 {0}
-#pack_dialog.cannot_write = 无法在目的文件夹创建压缩文件
-#unpack.unable_to_open_zip = 无法打开压缩文件 {0}
-#mkdir_dialog.title = 新建文件夹
-#edit_bookmarks_dialog.remove = 删除
-#mkdir_dialog.description = 创建文件夹
-#done = 完成
-#progress_dialog.hide = 隐藏
```
ahadas: s/unsed/unused in the commit message

ahadas:
very nice contribution!
please just fix the minor comment inside and I will merge it

src/main/resources/dictionary_ZH_CN.properties
mkortkamp: done

mkortkamp:
comment fixed

PR #23, opened 2016-04-16
support for retina displays

mkortkamp:
Changed macAppBundle configuration to include
NSHighResolutionCapable=true into application Info.plist. 
As a result we get nicer looking fonts on retina displays.


ahadas:
Thank you for this contribution!

PR #40, opened 2016-09-18
Prevent switching to discrete graphics on macOS

alexey-lysiuk:
JRE uses hardware acceleration by default leading to background activation of discrete graphics
the corresponding setting in application bundle stops this switching
integrated graphics is now used instead which helps to avoid redundant power drain
see https://developer.apple.com/library/content/qa/qa1734/_index.html


ahadas:
Thanks for this contribution!

PR #50, opened 2016-11-01
fix #45: function sendByEmail open bad tab

### LINKED ISSUE #45: function "send by email" open bad tab in preferences
When I want to send file by email, mucommander show info "You need to set your email parameters first". When I click OK, mucommader open preferences window but with bad tab(shortcuts instead mail).

jsebasct:
Arik,

Sorry for the delay, but now everything should be ok.


ahadas:
thanks for this contribution!

PR #95, opened 2017-04-22
Improvements to shell script gradlew

rdiez:
- Make it more robust, so that it stops on error.
  In the past, this script would blindly carry on after a command failed.
- Error on undefined variables, so that it is harder to
  make mistakes when maintaining the script.
- Warnings and errors are now output to stderr instead of stdout.
- This script can now handle more than 9 command-line arguments on Cygwin.

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -6,24 +6,34 @@
 ##
 ##############################################################################
 
+set -o errexit
+set -o nounset
+set -o pipefail
+
+
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
 DEFAULT_JVM_OPTS=""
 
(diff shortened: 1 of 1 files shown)


rdiez:
I'm sorry, I did not realise that the script was autogenerated. I filed a bug about those script issues in the Gradle repository. But we shouldn't merge any manual changes, so I am closing this pull request.

ahadas:
Well, I agree that we should try to avoid manual changes in auto-generated scripts and that the right thing to do in this case is making those changes in Gradle so they would be provided with future versions and for other projects as well. But I think that if we can achieve better functionality with minimal changes and minimal risk, we can do that. So let's see how Gradle's guys will prioritize the issues you open and decide what to do accordingly.

PR #97, opened 2017-05-01
Do not forward the 'home' and 'end' keys to the autocompletion pop-up list.

rdiez:
In previous versions of muCommander, the 'home' and 'end' keys were forwarded
to the pop-up list (if opened), which I found surprising. This version handles
the 'home' and 'end' keys always in the text field, like Firefox and others do.

### DIFF
--- a/src/main/java/com/mucommander/ui/autocomplete/TextFieldCompletion.java
+++ b/src/main/java/com/mucommander/ui/autocomplete/TextFieldCompletion.java
@@ -77,6 +77,11 @@ public class TextFieldCompletion extends CompletionType {
 
     	autocompletedtextComp.getDocument().addDocumentListener(documentListener);
 
+        // In previous versions of muCommander, the 'home' and 'end' keys were forwarded
+        // to the pop-up list (if opened), which I found surprising. This version handles
+        // the 'home' and 'end' keys always in the text field, like Firefox and others do.
+        boolean SHOULD_HOME_AND_END_KEYS_WORK_ON_POPUP_LIST = false;
+
         autocompletedtextComp.addKeyListener(new KeyAdapter() {
         	@Override
             public void keyPressed(KeyEvent keyEvent) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/mucommander/ui/autocomplete/TextFieldCompletion.java
```diff
@@ -77,6 +77,11 @@ public TextFieldCompletion(AutocompleterTextComponent comp, Completer completer)
 
     	autocompletedtextComp.getDocument().addDocumentListener(documentListener);
 
+        // In previous versions of muCommander, the 'home' and 'end' keys were forwarded
+        // to the pop-up list (if opened), which I found surprising. This version handles
+        // the 'home' and 'end' keys always in the text field, like Firefox and others do.
+        boolean SHOULD_HOME_AND_END_KEYS_WORK_ON_POPUP_LIST = false;
```
ahadas: This is not needed. if we don't allow the user to change that value (and I think we don't need to allow that), then let's not make it configurable because it makes the code below more cumbersome.

rdiez: About your first suggestion, what you propose is not intuitive. You may argue that this is completely subjective, but it is not the way Firefox, Chrome or Dolphin behave.

About your second point, if we do decide later to make it configurable, the flag is already there, so why throw away working code? You can remove it, if you wish, but please do not make me rework the pull request for such a trivial amendment, as it would take much longer than to modify the code in your editor and commit it!

ahadas: Right, it may be different in the applications you mentioned. I checked Firefox and it is indeed different. On the other hand, Eclipse (for instance) behaves just like muCommander does at the moment. So I don't think there is a clear standard that we violate. In my opinion, we should find what is most appropriate for muCommander.

Now, since we don't provide an easy way to jump to the bottom of the popup list (such as pressing the 'up' key when the first item is selected, like in Firefox), I think that dropping the support for jumping straight to the first and last items in the list would be a step back from user experience. That is subjective, of course.

I don't see why would it be more difficult for you to make the change. You can change the code, commit it to the branch of the pull request on your forked repository and this pull request will be updated. This is a similar process to what I would do.

rdiez:
OK, thanks for taking a look.

Issue #102, opened 2017-05-24
RFE: Single panel view on startup by default


christianrj:
Hi!

If possible, add an option in muCommander to set it to single panel view on startup.

Currently, if you click on "Toogle Single Panel View" and close the application, on the next start it will start on dual panel view.

Thank you!

PR #112, opened 2017-07-03
#102: Store/Restore Single Panel View state on startup

thomasuebel:
Hi,

This PR resolves RFE #102.
I have added Preferences to MuPreferences. The preference is stored when the ToggleUseSinglePanelAction is performed. It is restored during MainFrame init, when the action is performed depending on the toggle state.

There is no PullRequest template, so I hope this suffices.

On a side note: this is my first ever contribution to an Open Source project and I'm looking forward to your feedback. Thanks for your quick responses so far, and keeping this project alive! Being an avid user of muCommander myself - this was quite fun. :)

BR,
Thomas

### DIFF
--- a/src/main/java/com/mucommander/ui/main/MainFrame.java
+++ b/src/main/java/com/mucommander/ui/main/MainFrame.java
@@ -266,11 +266,6 @@ public class MainFrame extends JFrame implements LocationListener {
 
         // Set the custom FocusTraversalPolicy that manages focus for both FolderPanel and their sub components.
         setFocusTraversalPolicy(new CustomFocusTraversalPolicy());
-
-        // Restore single panel view state
-        if(MuConfigurations.getPreferences().getVariable(MuPreference.USE_SINGLE_PANEL_VIEW, MuPreferences.DEFAULT_USE_SINGLE_PANEL_VIEW)) {
-            ActionManager.performAction(ToggleUseSinglePanelAction.Descriptor.ACTION_ID, this);
-        }
     }
 
     public MainFrame(ConfFileTableTab leftTab, FileTableConfiguration leftTableConf,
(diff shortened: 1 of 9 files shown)


src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -260,11 +261,16 @@ public void windowClosing(WindowEvent e) {
...
         fireActivePanelChanged(activeTable.getFolderPanel());
 
-        // Set the custom FocusTraversalPolicy that manages focus for both FolderPanel and their subcomponents.
+        // Set the custom FocusTraversalPolicy that manages focus for both FolderPanel and their sub components.
         setFocusTraversalPolicy(new CustomFocusTraversalPolicy());
+
+        // Restore single panel view state
+        if(MuConfigurations.getPreferences().getVariable(MuPreference.USE_SINGLE_PANEL_VIEW, MuPreferences.DEFAULT_USE_SINGLE_PANEL_VIEW)) {
+            ActionManager.performAction(ToggleUseSinglePanelAction.Descriptor.ACTION_ID, this);
+        }
```
thomasuebel: Yeah, I don't use multiple windows either. I suppose, If I were to open a new window - I am effectively 'cloning' the current set up, right? If you open a new window, you'll also start at your currently open Folders. I would assume that, as a User, I'd also want to start a new windows with my currently selected view settings. 
I.e. when you select to use a single panel view, and start a new frame, you'd want to start with a single panel view as well.
I'm unsure about what would happen if you use the toggle in the second window, you opened, then close it - and then open a new window again. In the first window you'd have your original state, but the settings in this case would already have changed. It gives me a headache.

ahadas: Yeah, I agree with what you say and that it becomes complex :) So we can handle it separately. Let's try to extract this part to DefaultMainFramesBuilder so the configuration from snapshot.xml and preferences.xml will be applied on the main frames that are created on startup only for now.

thomasuebel:
Hi @ahadas ,

This pull request is now solely about the last status of the main frame. 

I've implemented the suggested changes: The state will be stored using the snapshot.xml. I've added the convenience method for the configuration key, which the DefaultMainFramesBuilder will make use of during createMainFrame(int).

I wasn't too sure if this part of the configuration is a WindowAttribute, but I've put it there for the time being - still open for your suggestions.

Thanks for steering me in the right direction.

Cheers!
-Thomas

ahadas:
@thomasuebel Thanks for this contribution. nice work!

PR #133, opened 2017-11-14
location-text-field component modified so as to have its text selected upon focus-gained event

rafic:
upon focus-gained event have text selected, so as to save the user the need to manually do so


ahadas:
Thank you for this contribution!

PR #158, opened 2018-06-10
Set up muCommander to build on Java 10 (though Proguard isn't supported)

holograph:
* Upgraded proguard to 6.0.3 (resolves JDK9+ linkage error)
* Upgraded launch4j to 2.4.4 (resolves build exception)
* Conditionally disabled proguard for JDK9+ (fails build because it looks for rt.jar)


ahadas:
Thanks!

PR #206, opened 2019-05-09
Update translations

olivierpaul:


ahadas:
Hi @olivierpaul , thanks for this second PR :)
I see that some changes got in to the translation files as part of the new plugin for docker images, but I would rather prefer to manage those translation files through [Zanata](https://translate.zanata.org/project/view/mucommander) and pull them from there. That being said, as I didn't set up the translations for version 0.9.4, I'll merge it and use it as baseline

PR #224, opened 2019-07-18
fix #218

### LINKED ISSUE #218: opening a specific directory via commandline opens two windows
If I call mucommander without any parameters:

```
mucommander
```
, it opens 1 window with the last opened directories, e.g. `/usr/share/doc/` and `$HOME`.

If I instead specify a certain directory to open:

```
mucommander /etc
```

it opens 1 window with the last opened directories (e.g. `/usr/share/doc` and `$HOME`), and a 2nd window with the specified directory (e.g. `/etc` and `$HOME`).

Related to that problem is, that the 2nd window i reduces to its minimum size, which in my case (Enlightenment filemanager under Debian Buster) is a very short titlebar with no actual visible window space.

The correct behaviour would be to open just the 2nd window, but that one in the correct size.

kousalik:
fixing #218 by skipping the snapshot windows recovery and using default frame size approach using the screen size

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/muCommander.java
+++ b/mucommander-core/src/main/java/com/mucommander/muCommander.java
@@ -24,6 +24,7 @@ import java.lang.reflect.Constructor;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


ahadas:
@kousalik could you please explain how did you pass arguments to mucommander on the master branch?

ahadas:
OK, so currently (since switching to OSGi) it's not possible to provide the initial location in the command line.  Open #227 to track this. This patch was verified on 0.9.3 and seems correct.

@kousalik do you use the [project's formatter](https://github.com/mucommander/mucommander/blob/master/mucommander-code-format.xml) in your editor?

ahadas:
@kousalik Thanks for this contribution!

ahadas:
Note: need to think of how to enable this flow with OSGi

PR #225, opened 2019-07-18
Issue #218 fix for the 0.9.3 stream

kousalik:
fixing #218 by skipping the snapshot windows recovery and using default frame size approach using the screen size


ahadas:
@kousalik this backport to the 0.9.3 branch is not needed, I don't think we will release a new version of 0.9.3 unless a critical bug is discovered.

PR #226, opened 2019-07-19
Improved build.gradle and documentation for easier and more flexible debugging

sandreas:
- Added check for environment variable `DEBUG` to enable remote debugging
- Improved `README.md` (headlines, section about pull requests, IDEA debugging example with screenshots and changed port to `5005`, which is the default setting)


ahadas:
@sandreas thank you for this contribution! this can really simplify debugging. 
The downside I see is in adding more text and screenshots to the readme page that should be minimal. I think the details about debugging through IntelliJ (and maybe the whole debugging section) should be separated out into a different wiki page, what do you think? could you pose a follow-up PR?

sandreas:
Well, of course we can move this into a wiki. I would follow your rules, since it is your project. 

At the moment I'm trying to get hot-swap-code working, but it seems that IntelliJ IDEA does not support this on OSGi containers. Perhaps I'll switch to eclipse, without hot-swap it is not really fun, resolving the drag & drop issue... so if switching to wiki, it would be great if you add some hints about your project setup - I'll add a real nice step by step guide then...

ahadas:
> Well, of course we can move this into a wiki. I would follow your rules, since it is your project.

It is considered to be best-practice to keep the readme page concise, with the most relevant information. Long and comprehensive readme page may make people get the wrong impression about the complexity of the project. I think the readme should tell "you can just clone the project and with a simple command (assuming Java is already installed), you can build and run the project from source", and link to other pages for more advanced things.

This is a community project, so please feel free to share your opinion :)
 
> At the moment I'm trying to get hot-swap-code working, but it seems that IntelliJ IDEA does not support this on OSGi containers. Perhaps I'll switch to eclipse, without hot-swap it is not really fun, resolving the drag & drop issue... so if switching to wiki, it would be great if you add some hints about your project setup - I'll add a real nice step by step guide then...

Actually, I'm not sure hot-swap would work with Eclipse, but worth a try.
Maybe, if we lose such capabilities, we should provide a way to package everything into a fat-jar, as we used to before, just for development and package into OSGi bundles only "for production". The downside, however, would be to maintain two different packaging strategies and less tests to the way the application is deployed for end-users.

sandreas:
> Actually, I'm not sure hot-swap would work with Eclipse, but worth a try.

Oh, this means you have to recompile on every change to test it? For me it takes about 30 Seconds to recompile...
I'm no Java / OSGi / gradle expert, but here is a writeup, that looks promising and could possibly speed up development - see chapter 4: 
https://sites.google.com/a/athaydes.com/renato-athaydes/posts/4freewaystohot-swapcodeonthejvm

> This is a community project, so please feel free to share your opinion!
Since I'm no Java expert - just interested in programming in general, a step by step dev environment setup would have helped me a lot to get into the project... I'll take care of it as soon as I think that I have a decent setup ;) At the moment it feels a bit exhausting to wait on the recompile everytime i changed the code :-) 

But lets not waste your time...
- If you know a better way than recompiling 30s for every change, I would love to see a short description what to do - it would be great to see how you setup your environment with eclipse in short detail
- If you find a way to apply the writeup above for hot swap, this would be great if it is an improvement
- If not, I'll try to fix the bug with my existing environment - but it will take some time to get into the code

Thank you for your explanations and politeness, I always like this kind of constructive discussions.

PR #229, opened 2019-07-24
Filter for shortcuts table

sandreas:
I implemented a text input filter for shortcuts - in my opinion it is useful, but I'm not sure, if the implementation is ok.

What do you think?

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/dialog/pref/general/ShortcutsPanel.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/dialog/pref/general/ShortcutsPanel.java
@@ -69,7 +69,7 @@ public class ShortcutsPanel extends PreferencesPanel {
     public ShortcutsPanel(PreferencesDialog parent) {
         super(parent, Translator.get("shortcuts_panel" + ".title"));
         initUI();
-        setPreferredSize(new Dimension(0, 0));
+        setPreferredSize(new Dimension(0,0));
 
         shortcutsTable.addDialogListener(parent);
     }
@@ -104,8 +104,8 @@ public class ShortcutsPanel extends PreferencesPanel {
      * Returns a panel that contains the shortcuts editor table.
      */
     private JPanel createCenterPanel() {
-        JPanel panel = new JPanel(new GridLayout(1, 0));
-        shortcutsTable.setPreferredColumnWidths(new double[]{0.6, 0.2, 0.2});
+        JPanel panel = new JPanel(new GridLayout(1,0));
+        shortcutsTable.setPreferredColumnWidths(new double[] {0.6, 0.2, 0.2});
         panel.add(new JScrollPane(shortcutsTable));
         return panel;
     }
@@ -129,7 +129,7 @@ public class ShortcutsPanel extends PreferencesPanel {
 
     private JPanel createButtonsPanel() {
         JPanel panel = new JPanel(new FlowLayout(FlowLayout.RIGHT));
-        panel.setBorder(BorderFactory.createEmptyBorder(0, 0, 0, 5));
+        panel.setBorder(BorderFactory.createEmptyBorder(0,0,0,5));
 
         RemoveButton removeButton = new RemoveButton();
 
@@ -151,9 +151,9 @@ public class ShortcutsPanel extends PreferencesPanel {
 
         private final Timer timer;
 
-        public DeferredDocumentListener(int timeOut, ActionListener listener, boolean repeats) {
+        public DeferredDocumentListener(int timeOut, ActionListener listener) {
             timer = new Timer(timeOut, listener);
-            timer.setRepeats(repeats);
+            timer.setRepeats(false);
         }
 
         public void start() {
@@ -190,7 +190,7 @@ public class ShortcutsPanel extends PreferencesPanel {
 
         final JComboBox<ActionCategory> combo = new JComboBox<>();
         combo.addItem(ActionCategory.ALL);
-        for (ActionCategory category : ActionProperties.getActionCategory())
+        for(ActionCategory category : ActionProperties.getActionCategory())
             combo.addItem(category);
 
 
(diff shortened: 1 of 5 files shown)


mucommander-core/src/main/java/com/mucommander/ui/dialog/pref/general/ShortcutsPanel.java
```diff
@@ -1,259 +1,335 @@
...
+                    filterField.requestFocus();
+                }
+            }
+        };
+
+
+        combo.addActionListener(filterActionListener);
+        combo.setSelectedIndex(0);
+
+        DeferredDocumentListener debounceListener = new DeferredDocumentListener(250, filterActionListener, true);
```
ahadas: I see you pass `true` here to set the time to repeated calls, why could it be repeated? shouldn't the trigger for filtering be when the document listener catches an event?
Also, the event that is dispatched from the event dispatching thread so what's the reason for using the timer? (that may be related to the first question above)

sandreas: Cool, thank you for the feedback...

Well, I'm not a java expert, but afaik the `setRepeats(true)` is only explicitly stating out that the event might be triggered more than once, because i think that the `focusGained` event can be emitted multiple times, even if the focus is already present for that element... 

Since the `filterField` loses focus on an update of the tableData (which was a tough problem to solve), it had to `requestFocus` again on every `actionPerformed`, if the source is `filterField`, so that you can still type after a table update. I'm not very proud of this solution, but it seems to work - a rework of the table update to not change the focus of the current element would have been a cleaner solution.

The timer is used, because if the table is updated on the fly, typing feels stuttering, at least on my system (1. typing, 2. table update, 3. requestFocus again - takes to long to feel natural). 250ms represents the median reaction time of a human, so i think the table update has a pretty good balance between "feels like on-the-fly" and "typing still feels natural".

I don't know, why my IDE thought formatting was a good idea, but if i find the time, i'll do the following things:

- remove the formatting completely - i don't think it is necessary
- check if `timer.setRepeats(true)` is really necessary and remove it if it is not...

After this i'll clean the pull request to only contain the changes required for this feature. Unfortunately it will take some time, because I'm busy at the moment.

ahadas: > Well, I'm not a java expert, but afaik the `setRepeats(true)` is only explicitly stating out that the event might be triggered more than once, because i think that the `focusGained` event can be emitted multiple times, even if the focus is already present for that element...

Ah ok, I see your point.
So yeah, without setting `setRepeats(true)` the timer only triggers one even, even calling `start/stop-start/restart` multiple times.
The downside though is that that we trigger an even every 250msec even when the text within the textfield widget is not changing, but we can live with that (the most important part is to cancel the timer when the focus is lost, as you did).
 
> Since the `filterField` loses focus on an update of the tableData (which was a tough problem to solve), it had to `requestFocus` again on every `actionPerformed`, if the source is `filterField`, so that you can still type after a table update. I'm not very proud of this solution, but it seems to work - a rework of the table update to not change the focus of the current element would have been a cleaner solution.

Interesting, did you find the reason the focus switched? is the table requests the focus?

> The timer is used, because if the table is updated on the fly, typing feels stuttering, at least on my system (1. typing, 2. table update, 3. requestFocus again - takes to long to feel natural). 250ms represents the median reaction time of a human, so i think the table update has a pretty good balance between "feels like on-the-fly" and "typing still feels natural".

Yeah, the resulting user experience is nice indeed.
I wonder if we can replace the time with a single thread as I did once for auto-completion - could you please [take a look at it](https://github.com/mucommander/mucommander/blob/master/mucommander-core/src/main/java/com/mucommander/ui/autocomplete/CompletionType.java#L188) and tell if this can be done similarly?
 
> I don't know, why my IDE thought formatting was a good idea, but if i find the time, i'll do the following things:
> 
> * remove the formatting completely - i don't think it is necessary

Well, the benefit of using a formatter is that styling is more consistent and we reduce/eliminate comment of type "please replace the TAB with 4 whitespaces" in the review. I this it would be better to import [this formatter](https://github.com/mucommander/mucommander/blob/master/mucommander-code-format.xml).

> * check if `timer.setRepeats(true)` is really necessary and remove it if it is not...
> 
> After this i'll clean the pull request to only contain the changes required for this feature. Unfortunately it will take some time, because I'm busy at the moment.

Cool :)

sandreas:
> Interesting, did you find the reason the focus switched? is the table requests the focus?

Not exactly... i think it has to do with this part:
```
// ShortcutsTable, method filter, ~ line 576
            ShortcutsTable.this.clearSelection();
            ((DefaultTableModel) getModel()).setRowCount(data.length);
            ShortcutsTable.this.repaint();
            ShortcutsTable.this.scrollToVisible(0, 0);

```
Things i did:
- I set the `setRepeats(false)` and it still works... so lets keep like this, as you suggested.
- Then I tried my very best with the formatting, but unfortunately IntelliJ is not able to apply all the eclipse formatting rules, even though there is a possibility to import an eclipse formatting specification (which i did).  I can't do anything else, I really hope it is acceptable now... perhaps you find the time to just reformat the piece of code - the review diff noise should be gone except whitespaces and these can be ignored on github...

mucommander-core/src/main/java/com/mucommander/ui/dialog/pref/general/ShortcutsPanel.java
sandreas: I think the things i could to are resolved. See comment below. If you aren't happy with these changes, i can't do anything about it, because my IDE is not able to do the code formatting for me like it is required.

ahadas: @sandreas ok, sorry - I thought this PR is still in the works. thanks for pinging. I'll rebase it soon and apply the formatter

ahadas:
@sandreas , thank you for this contribution, really nice work!

sandreas:
This is great, thank you. Glad i could help.

Issue #154, opened 2018-05-04
open command prompt 


systemsplanet:
This works under windows7 assigning to ALT-C

### commands.xml
```xml
<command alias="openCmd"  type="system" value="cmd /c start &quot;title&quot; /D &quot;$f&quot; cmd /k cd" display="Explorer"/>
```

### WinNtDesktopAdapter.java
   CommandManager.registerDefaultCommand(new Command(CommandManager.CMD_OPENER_ALIAS, FILE_OPENER_COMMAND, CommandType.SYSTEM_COMMAND, null));

### CommandManager.java
    public static final String                    CMD_OPENER_ALIAS              = "openCmd";

### DesktopManager.java 
```java
    public static final String                           OPEN_COMMAND_PROMPT  = "openCmd";

    innerRegisterOperation(OPEN_COMMAND_PROMPT, SYSTEM_OPERATION, new CommandOpenCommandPrompt());

    public static void openCommandPrompt(File file) throws IOException, UnsupportedOperationException {
        executeOperation(OPEN_COMMAND_PROMPT, new Object[] { file });
    }

    public static void openCommandPrompt(String file) throws IOException, UnsupportedOperationException {
        executeOperation(OPEN_COMMAND_PROMPT, new Object[] { file });
    }

    public static void openCommandPrompt(AbstractFile file) throws IOException, UnsupportedOperationException {
        executeOperation(OPEN_COMMAND_PROMPT, new Object[] { file });
    }

```

### CommandOpenCommandPrompt.java
```java
package com.mucommander.desktop;

import java.io.IOException;
import com.mucommander.command.Command;
import com.mucommander.command.CommandManager;
import com.mucommander.commons.file.AbstractFile;
import com.mucommander.process.ProcessRunner;

class CommandOpenCommandPrompt extends LocalFileOperation {
    @Override
    public boolean isAvailable() {
        return CommandManager.getCommandForAlias(CommandManager.CMD_OPENER_ALIAS) != null;
    }

    @Override
    public void execute(AbstractFile file) throws IOException {
        Command command;

        if ((command = CommandManager.getCommandForAlias(CommandManager.CMD_OPENER_ALIAS)) == null)
            throw new UnsupportedOperationException();

        ProcessRunner.execute(command.getTokens(file), file);
    }

    @Override
    public String getName() {
        Command command;

        if ((command = CommandManager.getCommandForAlias(CommandManager.CMD_OPENER_ALIAS)) != null)
            return command.getDisplayName();
[...]

PR #230, opened 2019-07-26
#154 open command prompt

sandreas:
Pull request für #154 with some improvements.

Only tested on Ubuntu 18.04 LTS with GNOME - perhaps some adjustments for the command line are required on other systems, but it should also work on Linux KDE, MacOS and Windows, etc..


sandreas:
Did a pull request on this one with some adjustments for Linux and MacOS, where opening a terminal should now also work.

See https://github.com/mucommander/mucommander/pull/230 for details.

PR #248, opened 2019-11-09
minor cleanup

ahadas:


ahadas:
@hajdam can you please take a look?

hajdam:
Sure, looks great. I knew there was probably better comparator possible, but I'm not that familiar with stream API yet. Good work.

ahadas:
> Sure, looks great. I knew there was probably better comparator possible, but I'm not that familiar with stream API yet. Good work.

thanks! :)

PR #261, opened 2020-01-08
Use bundled JRE as fallback on Windows

ahadas:


ahadas:
This seems to increase the boot time of the application in some cases.
As we're going to bundle the JRE with the application, dropping this.

PR #266, opened 2020-01-18
Remove copyright

ahadas:


sskras:
What is the reason to remove it from the source? Is there something like AUTHORS files present?

ahadas:
@sskras Hi, you meant to write it in #267, right? having an AUTHOR file doesn't seem to be a common practice in Github projects, as that information can be obtained using the Github interface and so that file becomes yet another resource to maintain. Do you think it is needed?

sskras:
No, I was interested just about the file headers.

I have no strong opinion about need to have it. It's just seemed natural (even if oldschool) to me to have contributors listed per source file. So they remain seen when another project copies it.

So I asked about the reason. I suppose, this style enables less updates / smaller diffs (due to static file headers).

ahadas:
@sskras Ah I see. The `@author` annotations at the file headers remain unchanged and state the contributors. Only the copyright line was dropped.

PR #376, opened 2020-09-27
Add ability to go next/previous image in ImageViewer

ahaltindis:
Hey,

Thank you for maintaining this great tool. I just want to add this simple but useful feature to ImageViewer. I also saw some commented out code for this feature so tried to uncomment them and follow same convention. Also I don't know if there is blocker for this but I couldn't find anything in the git history so I think whatever it is, it might not be valid anymore.

Thanks!


ahadas (approved):
Nice!
Thanks for this contribution

Issue #335, opened 2020-05-04
[Feature Request] calculate directory size by `F3`


sfuerte:
Default action for `F3` is a file `View`. Would be great if it calculated a directory size when the cursor is on one. 
A very missed feature comparing to Far Manager (https://www.farmanager.com/) and this clone (https://github.com/trol73/mucommander)

ahadas:
This is fixed in [v0.9.6](https://github.com/mucommander/mucommander/releases/tag/0.9.6-1)

PR #395, opened 2020-12-19
dir size calculation

sfuerte:
as per https://github.com/mucommander/mucommander/issues/335 and based on trolCommander's fork.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
@@ -51,10 +51,10 @@ abstract class AbstractViewerAction extends SelectedFileAction {
         super(mainFrame, properties);
 
         // Enable this action only if the currently selected file is not a directory and can be read.
-        setSelectedFileFilter(new AndFileFilter(
-            new FileOperationFilter(FileOperation.READ_FILE),
-            new AttributeFileFilter(FileAttribute.DIRECTORY, true)
-        ));
+        // setSelectedFileFilter(new AndFileFilter(
+        //     new FileOperationFilter(FileOperation.READ_FILE),
+        //     new AttributeFileFilter(FileAttribute.DIRECTORY, true)
+        // ));
     }
 
 
@@ -85,7 +85,7 @@ abstract class AbstractViewerAction extends SelectedFileAction {
                     catch(Exception e) {
                         InformationDialog.showErrorDialog(mainFrame);}
                 }
-                
+
                 // If it's distant, copies it locally before running the custom editor on it.
                 else {
                     ProgressDialog progressDialog = new ProgressDialog(mainFrame, Translator.get("copy_dialog.copying"));
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
@@ -51,10 +51,7 @@ abstract class AbstractViewerAction extends SelectedFileAction {
         super(mainFrame, properties);
 
         // Enable this action only if the currently selected file is not a directory and can be read.
-        // setSelectedFileFilter(new AndFileFilter(
-        //     new FileOperationFilter(FileOperation.READ_FILE),
-        //     new AttributeFileFilter(FileAttribute.DIRECTORY, true)
-        // ));
+        setSelectedFileFilter(new FileOperationFilter(FileOperation.READ_FILE));
     }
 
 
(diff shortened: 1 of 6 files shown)


mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
```diff
@@ -51,10 +51,10 @@ public AbstractViewerAction(MainFrame mainFrame, Map<String,Object> properties)
...
 
         // Enable this action only if the currently selected file is not a directory and can be read.
-        setSelectedFileFilter(new AndFileFilter(
-            new FileOperationFilter(FileOperation.READ_FILE),
-            new AttributeFileFilter(FileAttribute.DIRECTORY, true)
-        ));
+        // setSelectedFileFilter(new AndFileFilter(
+        //     new FileOperationFilter(FileOperation.READ_FILE),
+        //     new AttributeFileFilter(FileAttribute.DIRECTORY, true)
+        // ));
```
ahadas, line 57: I guess we want to change the filter to folders or files that are readable rather than disabling it (so that this action would still be disabled for non-folders that are non-readable), right?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/InternalEditAction.java
```diff
@@ -70,7 +72,13 @@ public InternalEditAction(MainFrame mainFrame, Map<String,Object> properties) {
...
     @Override
     protected void performInternalAction(AbstractFile file) {
-        EditorRegistrar.createEditorFrame(mainFrame, file, getIcon().getImage());
+        if (file.isDirectory()) {
+            FileSet fileSet = new FileSet();
+            fileSet.add(file);
+            new ChangePermissionsDialog(mainFrame, fileSet).showDialog();
+        } else {
+            EditorRegistrar.createEditorFrame(mainFrame, file, getIcon().getImage());
+        }
```
ahadas, line 81: don't we want this to happen also when custom-edit/view operation is defined? I'd say we should have the logic for folders in AbstractViewerAction#performAction. what do you think?

mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTableModel.java
```diff
@@ -93,18 +120,27 @@
...
      *
-     * @param compactSize true to use a compact size format, false for full size in bytes 
+     * @param compactSize true to use a compact size format, false for full size in bytes
      */
     static void setSizeFormat(boolean compactSize) {
         if(compactSize)
             sizeFormat = SizeFormat.DIGITS_MEDIUM | SizeFormat.UNIT_SHORT | SizeFormat.ROUND_TO_KB;
         else
-            sizeFormat = SizeFormat.DIGITS_FULL | SizeFormat.UNIT_NONE;
+        sizeFormat = SizeFormat.DIGITS_FULL;
```
ahadas, line 129: the indentation doesn't look right here, no?
and why changing that?

ahadas (commented):
looks ok overall
besides the comments inside, please drop a line in package/readme.txt

mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTableModel.java
sfuerte: my bad copy/pasting, was porting trolCommander's commit late night, will fix it.

mucommander-core/src/main/java/com/mucommander/ui/action/impl/InternalEditAction.java
sfuerte: hmm, but why? If someone overrides it with a custom action, why do we want to enforce some internal logic there?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
sfuerte: fixed.

mucommander-core/src/main/java/com/mucommander/ui/action/impl/InternalEditAction.java
ahadas: I think it depends on how the custom action is configured
if we use, let's say, Emacs as a viewer then there should be no problem to provide it with a folder
but that's a differently story in case we use something else that fails on folders

that said, it's not like someone lose functionality by defining a custom edit/view command - in that case, the edit/view actions on folders that we default to when no custom edit/view command is set can be invoked using their original shortcuts.

so I have no strong opinion on that, I'm ok with that if you think it's better that way

ahadas (approved):

Issue #403, opened 2021-01-26
No message shown on error while opening a file


essobedo:
Sometimes, when I try to open a file nothing occurs so I have no way to know what's wrong. The idea of this improvement is to provide the output message provided by the process launched.

essobedo:
The corresponding PR https://github.com/mucommander/mucommander/pull/404

essobedo:
Hi @ahadas, what do you think of this ticket and PR?

ahadas:
v0.9.7 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/0.9.7-1)

PR #404, opened 2021-01-26
Ref #403 - Provide output message on error

essobedo:
Fix for #403

## Motivation

No message is provided to the end-user when an error occurs while opening a file.

## Modifications:

* Change the signature of the methods to return a `CompletionStage<Optional<String>>` allowing to retrieve the output message as callback
* Add a specific `ProcessListener` to provide the output message through a callback
* Show the output message with a `InformationDialog.showErrorDialog` when it is possible


essobedo:
@ahadas Could you please review it 🙏  ? I would like to propose other improvements based on it 😄

ahadas:
@essobedo thanks for this contribution! looks great

PR #406, opened 2021-01-31
fix #405: Open the file with the viewer in case of error

### LINKED ISSUE #405: Open the file with the viewer in case of error
If we try to open a file that has an extension for which no editor has been affected, we get an error. The idea of this improvement is to give the opportunity to open the file with the viewer in case the file could not be opened natively.

essobedo:
Fix for #405 

## Motivation

Open the file even if there is no Editor defined for the file we try to open.

## Modifications:

* Adds a user preference to enable or not the ability to open the file with the viewer in case of an error
* Makes `OpenAction`, `OpenNativelyAction` and `TempExecJob` launch the viewer automatically to open the file in case of error if the user preference has been enabled
* Provides the new message in french and english


essobedo:
@ahadas what do you think of this PR and ticket?

olivierpaul (commented):
Hello @essobedo , when you want to display a single quote in a message ('), you need to put two single quotes in the properies file

essobedo:
@olivierpaul good catch, thx for the feedback

essobedo:
@olivierpaul fixed

mucommander-core/src/main/java/com/mucommander/ui/action/impl/OpenAction.java
```diff
@@ -35,13 +35,8 @@
...
 import com.mucommander.core.desktop.DesktopManager;
 import com.mucommander.job.impl.TempExecJob;
-import com.mucommander.process.ProcessRunner;
 import com.mucommander.text.Translator;
-import com.mucommander.ui.action.AbstractActionDescriptor;
-import com.mucommander.ui.action.ActionCategory;
-import com.mucommander.ui.action.ActionDescriptor;
-import com.mucommander.ui.action.ActionFactory;
-import com.mucommander.ui.action.MuAction;
+import com.mucommander.ui.action.*;
```
ahadas, line 39: please use the [formatter](https://github.com/mucommander/mucommander/blob/master/mucommander-code-format.xml) in your IDE. I don't mind how the imports are configured but when we all use the same formatter, it would reduce the number of changes on the PRs

ahadas (changes requested):
@essobedo thanks for this contribution and @olivierpaul  thanks for the review :)

besides using the formatter, please also drop a line about this change in the [readme file](https://github.com/mucommander/mucommander/blob/master/package/readme.txt#L35)

mucommander-core/src/main/java/com/mucommander/ui/action/impl/OpenAction.java
essobedo: @ahadas my bad, fixed

ahadas (approved):

ahadas:
thanks @essobedo !

essobedo:
@ahadas do you know when the next version will be released?

ahadas:
@essobedo hi, sorry - missed that question somehow.. I'd like to address what Google asked in order to certificate the plugin to Google Drive and that may take a bit of time. Let's see, if we'll have enough content, we can release the next version in a month or two even without it

essobedo:
@ahadas If you need help to address what Google asked, let me know. I would be happy to help if I can.

PR #411, opened 2021-02-05
Update dictionary_ko.properties

VenusGirl:
Added partial translation and correction of some mistranslations


ahadas (changes requested):
Thanks for this contribution @VenusGirl !

ahadas (approved):

PR #449, opened 2021-04-12
Fix FileViewer and FileEditor selection logic when a warning is raised

elemongw:
This fixes the FileViewer and FileEditor selection logic when a WarnUserException is raised.

I noticed this issue when trying to view large text files. Instead of displaying the file with the text viewer, the binary viewer was used instead.

I copied the logic to FileEditor as well since they share the same code.


ahadas (approved):
yeah, I've also noticed that - thanks for fixing this @elemongw !

PR #456, opened 2021-04-25
text-viewer: calculate line numbers panel width after file is loaded

elemongw:
With the current implementation, the initial display of a text file always has a width proportional to minimumDisplayDigits. The resulting visual issue is especially visible when opening files with lines in excess of 10000. With this commit, the width is calculated after the file is loaded and as a result it is set appropriately.
 
I have also added a second commit that increases the padding of the line numbers in the panel for better display, let me know how you feel about it.

EDIT: Added separate commit with fix for remembering panel visibility status.


ahadas (approved):
looks great, thanks!

PR #477, opened 2021-05-18
drop inbuilt tar format provider and use Apache Commons Compress instead

### LINKED ISSUE #452: Support for tar files in the pax format
The inbuilt provider for tar files is missing support for the posix format as defined by the POSIX.1-2001 specification (pax format). Currently, when an archive in this format is opened by muCommander, additional PaxHeaders.* folders will be be displayed which contain files with the metadata that this format supports.

On the official documentation of GNU tar, there is information that this will be the default in a future release (current default is gnu format). This note seems to have been there for a while so "soon" is relative but at the least openSUSE has already set it as the default output file format which is how I came up with the issue (and seems to have had it for quite some time, I faced this issue even on some quite old archives). 

The latest commons-compress (which is a continuation of the tar code from ant that muCommander also uses) appears to have good-enough support for this specification. During some tests I run, TarArchiveEntry.getNextTarEntry (from commons-compress) would return the correct file list while TarEntry.getNextEntry (from mucommander) included the PaxHeaders folders and files in the resulting file list.

The pax support code could potentially be ported from commons-compress or mucommander-format-tar could be ported to use the upstream project directly. I saw there is an entry to improve handling of archive files on the roadmap so I guess there may already be plans affecting this. If not, what would be the prefered way of resolving this?

ref: https://www.gnu.org/software/tar/manual/html_section/Formats.html

elemongw:
Closes #452

This pull request drops the inbuilt tar format provider and replaces it with upstream Apache Commons Compress. Apart from now supporting tar files in the pax format as mentioned on the linked ticket, this should also have the added bonus of incorporating features/bugs fixed in the upstream project over the years (from what I found out while implementing this, it seems the code in muCommander had been forked a long time ago from ant; commons-compress has the same base and continued development). 

I read that the next version of commons-compress adds support for random read access for tar files so once it is released I will take a look into supporting it, similarly to how it is done for zip files, which should considerably improve performance on some situations.

As a final note, the inbuilt tar provider used com.mucommander.commons.io.bufferpool in some places. This will no longer be the case now that commons-compress is used but I could not find any indication of it making much of a difference.


ahadas (commented):
looks great!
one minor comment inside and could you please drop a line in the [readme file](https://github.com/mucommander/mucommander/blob/master/package/readme.txt)?

ahadas:
Ah and one more thing - can you please change the commit message to start with `fix #452:` so it would be easy to jump from the change to the issue later on?

elemongw:
Of course! All done, let me know if it looks ok now.

PR #484, opened 2021-05-31
skip method of InputStream should return 0 when EOF is reached

elemongw:
As the commit message says, the skip method of an InputStream should return 0 when EOF is reached. I hit a bug while running the nightly where muCommander would get stuck when trying to open some tar files. The issue ended up being that commons-compress depends on this behaviour and due to -1 being returned it would get stuck in an endless loop

### DIFF
--- a/mucommander-commons-io/src/main/java/com/mucommander/commons/io/RandomAccessInputStream.java
+++ b/mucommander-commons-io/src/main/java/com/mucommander/commons/io/RandomAccessInputStream.java
@@ -95,10 +95,10 @@ public abstract class RandomAccessInputStream extends InputStream implements Ran
      * plus the number of bytes to skip doesn't exceed the length of this stream as returned by {@link #getLength()}.
      * If it does, all the remaining bytes will be skipped so that the offset of this stream will be positioned to
      * {@link #getLength()}.
-     * Returns <code>-1</code> if the offset is already positioned to the end of the stream when this method is called.
+     * Returns <code>0</code> if the offset is already positioned to the end of the stream when this method is called.
      *
      * @param n number of bytes to skip
-     * @return the number of bytes that have effectively been skipped, -1 if the offset is already positioned to the
+     * @return the number of bytes that have effectively been skipped, 0 if the offset is already positioned to the
      * end of the stream when this method is called
      * @throws IOException if something went wrong
      */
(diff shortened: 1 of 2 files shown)


mucommander-commons-io/src/main/java/com/mucommander/commons/io/RandomAccessInputStream.java
```diff
@@ -95,10 +95,10 @@ public void readFully(byte b[], int off, int len) throws IOException {
      * plus the number of bytes to skip doesn't exceed the length of this stream as returned by {@link #getLength()}.
      * If it does, all the remaining bytes will be skipped so that the offset of this stream will be positioned to
      * {@link #getLength()}.
-     * Returns <code>-1</code> if the offset is already positioned to the end of the stream when this method is called.
+     * Returns <code>0</code> if the offset is already positioned to the end of the stream when this method is called.
      *
      * @param n number of bytes to skip
-     * @return the number of bytes that have effectively been skipped, -1 if the offset is already positioned to the
+     * @return the number of bytes that have effectively been skipped, 0 if the offset is already positioned to the
```
ahadas, line 101: I see we rely on the negative returned value in `CompoundInputStream` and `StreamUtils#skipFully`. The former is unused as far as I can tell. It's harder to track the implication of changing the latter. Need to take a better look to see it's safe

ahadas: I removed the whole `com.mucommander.commons.io.compound` package (that contained `CompoundInputStream`) [here](https://github.com/mucommander/mucommander/commit/f721457aa6defca5589e0a80765f909a75df2385)

elemongw: You are right, I totally missed that. Looks like changing StreamUtils#skipFully to check if nbSkipped==0 should be enough, right? Let me know once you take a better look and I'll fix it.

ahadas: Yeah, this method can't be called with n==0 from `StreamUtils#skipFully` so if this method returns 0 to the call from `StreamUtils#skipFully` it means that we couldn't skip anything because we're at the end of the steam so this change looks good to me, please change and I'll get it in

ahadas (approved):
thanks!

PR #490, opened 2021-06-04
store preferences in installation directory when in portable mode

elemongw:
This affects the portable distribution (by default).

In order to detect whether the application should start in portable mode, an empty file with the name `.portable` is used. This file has to reside in the installation directory.

I chose the name and location of the portable preferences directory based on the information available on the wiki [1]. Considering there is a conf folder now, `conf/mucommander` sounds like a good option as well (although I also like the first option of having the preferences directory as top-level in the installation directory instead of inside a folder that is shipped with the distribution).

This PR changes the previous behaviour so it may be useful for some users (if anyone likes to use the portable distribution with global settings) to note that they can delete the `.portable` file from the installation directory in order to revert to the previous behaviour. If you think this would be useful, I could maybe add a line about in the readme.txt.

[1] https://github.com/mucommander/mucommander/wiki/Preference-Files

### DIFF
--- a/package/readme.txt
+++ b/package/readme.txt
@@ -41,6 +41,7 @@ Improvements:
 - Tar archives with headers in the pax format are now supported.
 - The free space indicator in the status bar is updated right after relevant events like switching to the application.
 - Symbolic links are unpacked from tar archives.
+- The portable version stores preferences files in the installation directory.
 
 Localization:
 - Updated Russian translation
--- a/src/main/java/com/mucommander/main/muCommander.java
+++ b/src/main/java/com/mucommander/main/muCommander.java
@@ -291,18 +291,26 @@ public class muCommander
         muCommander.copySystemProperties(configProps);
 
         File preferencesFolder;
-        if (configuration.preferences == null) {
-            try {
-                preferencesFolder = UserPreferencesDir.getDefaultPreferencesFolder();
-            } catch(RuntimeException e) {
-                System.err.println("Failed to retrieve default preferences folder: " + e.getMessage());
-                return;
-            }
-        } else {
+        if (configuration.preferences != null) {
             try {
                 preferencesFolder = UserPreferencesDir.getPreferencesFolder(configuration.preferences);
             } catch(RuntimeException e) {
                 System.err.println("Failed to retrieve specified preferences folder: " + configuration.preferences);
+                return;
+            }
+        } else if (new File(codeParentFolder, ".portable").exists()) {
+            String portableDir = new File(codeParentFolder, ".mucommander").getAbsolutePath();
+            try {
+                preferencesFolder = UserPreferencesDir.getPreferencesFolder(portableDir);
+            } catch(RuntimeException e) {
+                System.err.println("Failed to retrieve portable preferences folder: " + portableDir);
+                return;
+            }
+        } else {
+            try {
+                preferencesFolder = UserPreferencesDir.getDefaultPreferencesFolder();
+            } catch(RuntimeException e) {
+                System.err.println("Failed to retrieve default preferences folder: " + e.getMessage());
                 return;
             }
         }
--- a/build.gradle
+++ b/build.gradle
@@ -411,7 +411,8 @@ task portable(dependsOn: [createExe, createBundlesDir], type: Zip) {
     from ('package') { include 'license.txt', 'readme.txt' }
     def tempDir = File.createTempDir()
     tempDir.deleteOnExit()
-    from ({ new File(temporaryDir, 'felix-cache').mkdirs(); temporaryDir }) {
+    from ({ new File(temporaryDir, '.portable').createNewFile(); new File(temporaryDir, 'felix-cache').mkdirs(); temporaryDir }) {
+        include '.portable'
         include 'felix-cache'
         dirMode 0777
     }


src/main/java/com/mucommander/main/muCommander.java
```diff
@@ -291,18 +291,25 @@ public static void main(String[] args) throws Exception
...
-        if (configuration.preferences == null) {
+        if (configuration.preferences != null) {
             try {
-                preferencesFolder = UserPreferencesDir.getDefaultPreferencesFolder();
+                preferencesFolder = UserPreferencesDir.getPreferencesFolder(configuration.preferences);
+            } catch(RuntimeException e) {
+                System.err.println("Failed to retrieve specified preferences folder: " + configuration.preferences);
+                return;
+            }
+        } else if (new File(codeParentFolder, ".portable").exists()) {
```
ahadas, line 301: Any reason for not creating the `.mucommander` folder within the installation directory instead of having the `.portable` file?

ahadas (commented):
Yeah, the portable version should store the preferences files within the installation directory by default, thanks for posting this!

ahadas:
@elemongw note that I merged [this commit](https://github.com/mucommander/mucommander/commit/538fde24437e529b065dc2423ff876aad3014dca) to create the `felix-cache` folder within the archive without adding that folder to the repo. It can be used as a reference here and it lead to a conflict with this PR that needs to be resolved

package/readme.txt
```diff
@@ -41,6 +41,7 @@ Improvements:
 - Tar archives with headers in the pax format are now supported.
 - The free space indicator in the status bar is updated right after relevant events like switching to the application.
 - Symbolic links are unpacked from tar archives.
+- Preferences are stored in the installation directory when in portable mode.
```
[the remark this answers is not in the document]
elemongw: You are right, it is clearer worded that way. I changed it.

src/main/java/com/mucommander/main/muCommander.java
```diff
@@ -291,18 +291,25 @@ public static void main(String[] args) throws Exception
...
-                System.err.println("Failed to retrieve default preferences folder: " + e.getMessage());
+                System.err.println("Failed to retrieve portable preferences folder: " + portableDir);
                 return;
             }
         } else {
             try {
-                preferencesFolder = UserPreferencesDir.getPreferencesFolder(configuration.preferences);
+                preferencesFolder = UserPreferencesDir.getDefaultPreferencesFolder();
             } catch(RuntimeException e) {
-                System.err.println("Failed to retrieve specified preferences folder: " + configuration.preferences);
```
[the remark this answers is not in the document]
elemongw: Didn't mean to remove that and not sure how I missed it. My bad.

src/main/java/com/mucommander/main/muCommander.java
elemongw: I prefer having a clear distinction (the .portable file) of when to switch to portable mode, with the thinking that there may be more than one things being changed in the behaviour of the application (although in this case and for now there is only one thing being changed) and the .portable file is a kind of a flag for switching all of them. I also like it better being able to switch mode by removing/creating the .portable file instead of renaming the .mucommander folder. This is a personal preference though and I have seen it done both ways. If you think it would be better to have the .mucommander folder included in the distribution instead let me know and I will update the PR.

elemongw:
I was finally able to continue working on this one. I replied on the individual comments and updated the PR. I also used your commit as a reference for including the .portable file in the portable distribution without keeping it in the git repo.

I believe the only folder written in a non-portable location now is felix-cache, right? I am not sure if this matters, I might need to check how temporary files are handled by other applications when in portable mode. Do you have a preference on this?

src/main/java/com/mucommander/main/muCommander.java
ahadas: I thought it would be better to create the `.mucommamnder` folder with the right permissions so different users would be able to execute the same portable app (imagine the application is first started by root so the `.mucommander` is set with permissions for the root user and so we won't be able to start the app later on by a non-root user). But even if the `.mucommander` folder is set with `777`, the files inside this folder would be created by the user that was the first to initiate the application so we'll have the same problem, I think...

ahadas: So maybe we should "remember" that we run in a portable mode and set the permission on the preferences file accordingly? what do you think?

elemongw: That makes sence and it would be very useful for your described scenario. If we proceed with that, I agree that remembering that we run in portable mode is how I feel it would be best to implement it as well. I am not sure if I like it as the default behaviour though. If I am not mistaken, in a multiuser environment with a user having muCommander in their home directory, this would allow other users to run it (I believe read permissions are usually allowed by default) and overwrite their settings, right? I think as a user I would not expect this to be allowed by default and although it would take a malicious user to cause issues I would still like to safeguard against it. In a situation where a user does want this to be allowed, they could manually set the permissions of the .mucommander folder to 777.

As a side note regarding .mucommander vs .portable, another scenario just crossed my mind. To update a portable installation, a user would probably copy the contents of the new version over the contents of the older installation. If they use macOS and Finder to do that, and chooses to replace/merge all, they would lose all their settings as the empty .mucommander folder from the new version would replace the old one which contained the user settings. Then again this could be considered a too specific scenario and a user error (and a bad default by Finder) but I also have fallen to this trap before myself and replaced the contents of folders I didn't mean to.

In any case, I do not feel strongly about any of the above and I am not certain about some of the assumptions I made. What do you think, are these valid concerns or are they too specific scenarios?

ahadas: Yeah, those are valid arguments/concerns.
I think it really depends on the use case.
Generally, if I copy a file to, let's say a USB drive, and plug it elsewhere - I'll have access to the copied file also as a different user. I think it was that way also when mucommander stored the configuration files within the app folder for portable versions so let's go with that.

ahadas (approved):
Thanks!

elemongw:
Regarding felix-cache, do you think we should do anything about it?

ahadas:
> Regarding felix-cache, do you think we should do anything about it?

I think the handling of the felix-cache folder is ok the way it is - we better keep it in the temporary folder (had it been possible I would have dropped it altogether) than storing it within the installation folder. It's not something you want to "take with you" when moving between machines and it will just consume more space on the shareable device

Issue #463, opened 2021-05-09
muCommander unresponsive when searching


Andreas0602:
muCommander 0.9.7 on macOS Big Sur becomes unresponsive when searching. See the high CPU load in the attached screenshot. Also note the high memory usage of 16 GB.

<img width="528" alt="Bildschirmfoto 2021-05-09 um 11 05 32" src="https://user-images.githubusercontent.com/79788982/117566457-0091f000-b0b7-11eb-958b-f19c222093e2.png">

**Steps to reproduce:**
1. Go to your home directory.
2. File -> Find ...
3. Search for filename "muCommander", case insensitive, include subfolders, unlimited search depth, no text search
4. Start the search and wait ~ 30 seconds

Andreas0602:
I've made some more experiments to find out which folder in my home directory is causing the problem. It's the folder ~/Library/Containers/. Each sub-folder in ~/Library/Containers/ contains another sub-folder named "Data", which in turn contains several symbolic links to folders in my home directory, e.g. to the folder "Music". I guess these symbolic links have something to do with the problem. muCommander seems to follow these symbolic links during the search.

By the way, Double Commander has an option "Follow symlinks" in the search dialog. When that option is turned off, searching the folder ~/Library/Containers/ is very fast, and when it's turned on, searching takes a very long time. But in contrast to muCommander, CPU load and memory usage remain normal all the time.

ahadas:
Thanks for the further investigation @Andreas0602 
The CPU consumption can be decrease by limiting the number of threads that will be used during the search, the question is what should be the default - maybe a pool of two threads would be a good default.
Symbolic links - they sure make the search look at more places but as long as they are not cyclic I think it shouldn't take significantly more time to examine them. That said, adding an option of "Follow symlinks" makes sense.
As for the memory, I guess that would require a deeper look at our caches..

Andreas0602:
I don't think that there are cyclic links. When I open a terminal and enter the below command, the generated file ls.txt will have a size of about 1 GB. That's very large, but it's not infinitely large.

`ls -RL ~/Library/Containers > ~/Downloads/ls.txt 2>/dev/null`

So, there are "only" two things that need to be improved:
- The search function needs to get better at handling large directory structures.
- The search dialog needs an option "Follow symlinks".

I guess the first bullet is more difficult to implement than the second. But it's not impossible, as Double Commander, ForkLift, and other file managers show.

PR #495, opened 2021-06-26
Fix handling of search strings that resemble a path to an archive file

ahadas:


ahadas:
Thanks for reporting this, it also gave me a reason to make some code cleanup I planned to do related to file search :)

Andreas0602:
Thank you very much for this fix, @ahadas! By the way, fyi: this fix does **not** fix the issue #463. It would be great if you could fix that too :)

PR #506, opened 2021-09-20
add support for the xz file format and other enhancements 

elemongw:


mucommander-format-xz/build.gradle
```diff
@@ -0,0 +1,29 @@
+// In this section you declare where to find the dependencies of your project
+repositories.jcenter()
+
+dependencies {
+    api project(':mucommander-commons-file')
+
+    implementation 'org.osgi:osgi.core:7.0.0'
+    implementation 'org.tukaani:xz:1.9'
```
ahadas, line 8: any reason not to use the wrappers in apache commons-compress (that we already use) instead?

package/readme.txt
```diff
@@ -52,6 +52,7 @@ Bug fixes:
 - Restore file table configuration (column visibility, width and position) on startup also when custom folders are set.
 - The default keyboard shortcut for packing files (CTRL+I) opens the 'Pack files' dialog on macOS.
 - File search works with search strings that resemble a path to an archive file, e.g., find:///home/*.zip.
+- Improve the handling of edge cases when opening compressed archives and choosing a custom extension.
```
ahadas, line 55: this needs to be more informative or we should rather consider this part a code change and drop it - what do you think?

mucommander-format-xz/src/main/java/com/mucommander/commons/file/archive/xz/XzArchiveFile.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * XzArchiveFile provides read-only access to archives in the xz format.
+ *
+ * <p>The actual decompression work is performed by the {@link org.tukaani.xz.XZInputStream} class.</p>
+ *
+ * @see com.mucommander.commons.file.archive.xz.XzFormatProvider
+ * @author Giorgos Retsinas
+ */
+public class XzArchiveFile extends AbstractROArchiveFile {
```
ahadas, line 41: I see [here](https://commons.apache.org/proper/commons-compress/javadocs/api-1.6/org/apache/commons/compress/compressors/xz/package-summary.html) that both input and output streams exist for XZ, why making it read-only?

ahadas (changes requested):
@elemongw thanks for submitting this!

mucommander-format-xz/build.gradle
elemongw: The XZ support in commons-compress is a short wrapper around xz-java (and requires it to work) so I didn't see much benefit in using it and went with using xz-java directly. Let me know if you feel otherwise.

package/readme.txt
elemongw: Initially I added this as there are user-facing changes, some bugs were fixed and some things work better now. From the top of my head an example would be Open as -> tar.bz2 failing to work properly or the filename chosen not being very good when the file extension was wrong. However, I could not come up with a good summary and I don't like it much either so I'm for dropping it unless you can come up with something better.

mucommander-format-xz/src/main/java/com/mucommander/commons/file/archive/xz/XzArchiveFile.java
elemongw: I think we will need to rework RWArchiveEntryFile in order to do this. Currently, when a file inside an archive is updated it removes the existing entry and then adds back the updated one. However, this does not make sense for single-file formats like XZ. The same is true for AbstractRWArchiveFile and its methods addEntry and deleteEntry. I can take a deeper look into this eventually and then also update gzip and bzip2 format support which currently work the same way.

In the meantime, I added support for compressing files using XZ in the archiver which did not require any significant changes.

package/readme.txt
```diff
@@ -32,7 +32,7 @@ What's new since v0.9.7 ?
 -----------------------
 
 New features:
--
+- Support opening compressed files in the XZ file format.
```
ahadas, line 35: this can now indicate that compression is also supported with something like "Support compression and decompression of XZ archives"

package/readme.txt
ahadas: agree, lets drop it then

mucommander-format-xz/build.gradle
ahadas: I took a deeper look and saw that xz-java is not embedded in commons-compress as I thought but is rather set as an optional dependency of commons-compress. I wanted to avoid having the xz-java classes twice but as it's not the case, that's fine

ahadas (changes requested):

package/readme.txt
elemongw: Done

package/readme.txt
elemongw: Updated

ahadas (approved):

ahadas:
Thanks!

PR #513, opened 2021-10-01
fix binary viewer not closing InputStream

elemongw:


ahadas (approved):
Thanks!
CC @hajdam

hajdam:
Thanks and sorry. Plan is to rework it, but it's still my fault I left it like that :-(

Issue #101, opened 2017-05-24
RFE: Add support to SMB v2


dylan666:
muCommander uses SBM V1, isn't it?
Please add support to SMB v2 and an option to choose V1 or V2

ahadas:
@dylan666 , right. That's because JCIFS supports only v1 and as far as I understand have no intention to support later versions. Is there a specific feature in v2 that you need? Do you know of any Java library that supports v2?

Destynar:
Due to Wannacry, I disabled SMB1 on my servers completly. And I am looking for a programm to add SMB2 functionality to some stray WinXPs that are still in use. Alas, the only app I coud find that uses its own SMB is this one, and it's still SMB1. And java-based...
I am certainly looking forward to the adding of SMB2+ functionality.

MichaelMasuch:
Would be very thankful if SMB v2 (v3) was supported

ahadas:
@MichaelMasuch Thanks for raising this again. Seems [there are Java clients for v2 and v3 that are worth checking:](https://www.jcifs.org) jcifs-ng and smbj

Neustradamus:
Any news on it?

dylan666:
Any news?

ahadas:
Version 1.1.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.1.0-1)

Neustradamus:
@ahadas: Nice, good job, thanks!

PR #522, opened 2021-10-21
WIP: Basic SMBv2 support through jcifs-ng library

oujesky:
This should eventually resolve issue #101

The `jcifs-ng` library (https://github.com/AgNO3/jcifs-ng) which supports SMBv2+ protocol is almost a drop-in replacement for the original `jcifs`.

The PR is not final yet, as I have very little knowledge of SMB internals and there are some parts that needs further rewrite in the `mucommander-protocol-smb` module (marked with `@todo`). Any help from the community is more than welcome. 

In my environment (connecting to a Synology DiskStation running DSM 7.0) I can access the files, copy, move etc. but the main observation so far is that it's significantly slower than previously when connecting using SMBv1

### DIFF
--- a/mucommander-protocol-smb/build.gradle
+++ b/mucommander-protocol-smb/build.gradle
@@ -5,7 +5,7 @@ dependencies {
     api project(':mucommander-protocol-api')
     api project(':mucommander-translator')
 
-    implementation 'eu.agno3.jcifs:jcifs-ng:2.1.6'
+    implementation 'eu.agno3.jcifs:jcifs-ng:2.1.7'
 
     // Use JUnit test framework
     testImplementation 'junit:junit:4.12'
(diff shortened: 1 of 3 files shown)


ahadas:
@oujesky sorry for having it waiting for so long, I somehow missed that PR (I saw it briefly when looking at #101) and thanks for posting this. I'll review it soon

ahadas:
@oujesky first, sorry for the time it took me to get back to this and thanks again for posting that. I've addressed the TODOs inside and it seems to work great.

I compared the performance on my env - I manage to copy to a remote Windows machine 5 times faster than with the previous library and to copy from that machine 10 times faster. I don't see any difference between v2.1.6 and v2.1.7 and I can't recall any recent change that might explain the different between what I see on my environment and what you described that happened on your environment. can you please describe in more details the test you did? since it's been a while, would it make sense to try again with the latest version?

PR #530, opened 2021-12-16
Add Publisher and Version to NSI script

Trenly:


ahadas:
Yes, this seems to work properly
Thanks!

ahadas (approved):

Issue #545, opened 2022-01-02
Search with less Keystrokes


gerolfscherr:
when I am searching for a keyword in one or several files, I have to press F3 to view the file, then ctrl-f to enter the search term, then F3 again to 'search next', then close the file and open another one and the process begins.

A much better strategy would be to change the behavour a little bit:
- F3 - if no search keyword was entered, open search dialog, else 'search again'
- make the search keyword permanent in memory, so it is retained between files/searches

So when I am searching for a keyword in a couple of files (which is a task I do quite often while programming), it just comes down to repeated presses of F3 and Esc:

F3: view file
F3: the search dialog opens and I enter a keyword.
F3 (repeatedly): browse through the other occurrences.
Esc - close file.
(navigate to other file)
F3, F3, F3, F3... view and find first occurrence / subsequent occurrences
rinse and repeat.

gerolfscherr:
https://github.com/mucommander/mucommander/pull/546 this would fix the issue.

ahadas:
Version 1.0.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.0.0-1)

PR #546, opened 2022-01-02
Search with less Keystrokes #545

gerolfscherr:


ahadas:
@gerolfscherr thanks for this contribution! I'll polish it a bit in a separate PR

PR #549, opened 2022-01-03
Search for size

gerolfscherr:
+ Search for File Size (greater, less, equals) with Units (bytes, kB, MB, GB)

+ Small tweaks for the Search dialog which should enhance productivity:
- '*' as default file name, select-all-on-focus enabled. 
- mnemonic (default:alt-t) for search text.
- as a result the user (me) can search using fewer keystrokes: If I want to search for file name I can just start typing, but if I want to search for file content (text) I type alt-t and enter the search text.

TODO: I am not too proud of the layout in the dialog.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
+++ b/mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
@@ -36,6 +36,7 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JSpinner;
 import javax.swing.JTextField;
+import javax.swing.JComboBox;
 import javax.swing.SpinnerNumberModel;
 import javax.swing.border.Border;
 import javax.swing.event.DocumentEvent;
@@ -46,6 +47,7 @@ import com.mucommander.commons.file.FileFactory;
 import com.mucommander.commons.file.FileURL;
 import com.mucommander.commons.file.util.PathUtils;
 import com.mucommander.commons.util.Pair;
+import com.mucommander.commons.util.StringUtils;
 import com.mucommander.commons.util.ui.dialog.DialogToolkit;
 import com.mucommander.commons.util.ui.dialog.FocusDialog;
 import com.mucommander.commons.util.ui.layout.ProportionalGridPanel;
@@ -55,7 +57,11 @@ import com.mucommander.commons.util.ui.spinner.IntEditor;
 import com.mucommander.text.Translator;
 import com.mucommander.ui.action.ActionProperties;
 import com.mucommander.ui.action.impl.FindAction;
+import com.mucommander.ui.dialog.InformationDialog;
 import com.mucommander.ui.main.MainFrame;
+import com.mucommander.ui.text.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Dialog used to set parameters for file searching.
@@ -63,6 +69,9 @@ import com.mucommander.ui.main.MainFrame;
  * @author Arik Hadas
  */
 public class SearchDialog extends FocusDialog implements ActionListener, DocumentListener {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SearchDialog.class);
+
     private JTextField searchFilesField;
     private JTextField searchInField;
     private JCheckBox searchInSubfolders;
@@ -82,6 +91,11 @@ public class SearchDialog extends FocusDialog implements ActionListener, Documen
     private JCheckBox textCase;
     private JCheckBox textRegex;
 
+    private JComboBox<SizeUnit> cboSizeUnit = new JComboBox(SizeUnit.VALUES);
+    private JComboBox<SizeRelation> cboSizeRel = new JComboBox(SizeRelation.VALUES);
+    private JTextField tfSize = new EnhancedTextField(8, true);
+
+
     // Store last values, initialized to the default values
     private static boolean lastSearchInSubfolders = true;
     private static boolean lastSearchInArchives = false;
@@ -98,6 +112,10 @@ public class SearchDialog extends FocusDialog implements ActionListener, Documen
     private static String lastText = "";
     private static boolean lastTextCase = true;
     private static boolean lastTextRegex = false;
+    private static Long lastSize = null;
+    private static SizeRelation lastSizeRel = SizeRelation.VALUES[0];
+    private static SizeUnit lastSizeUnit = SizeUnit.VALUES[0];
+
 
     private JButton searchButton;
     private JButton cancelButton;
@@ -127,7 +145,8 @@ public class SearchDialog extends FocusDialog implements ActionListener, Documen
         XAlignedComponentPanel compPanel = new XAlignedComponentPanel(5);
 
         String searchFor = PathUtils.removeLeadingSeparator(searchURL.getPath());
-        searchFilesField = new JTextField(searchFor);
+        if (searchFor.isEmpty()) searchFor = "*";
+        searchFilesField = new EnhancedTextField(searchFor, true);
         searchFilesField.getDocument().addDocumentListener(this);
         compPanel.addRow(Translator.get("search_dialog.search_files"), searchFilesField, 5);
 
@@ -217,8 +236,9 @@ public class SearchDialog extends FocusDialog implements ActionListener, Documen
         textSearchPanel.setBorder(BorderFactory.createTitledBorder(Translator.get("Text search (Optional)")));
         compPanel = new XAlignedComponentPanel(5);
 
-        searchTextField = new JTextField(lastText);
-        compPanel.addRow(Translator.get("search_dialog.search_text"), searchTextField, 10);
+        searchTextField = new EnhancedTextField(lastText, true);
+        JLabel l = compPanel.addRow(Translator.get("search_dialog.search_text"), searchTextField, 10);
+        l.setDisplayedMnemonic(Translator.get("search_dialog.search_memonic", "T").charAt(0));
 
         textCase = new JCheckBox(Translator.get("search_dialog.text_case_sensitive"), lastTextCase);
         compPanel.addRow("", textCase, 10);
@@ -234,6 +254,8 @@ public class SearchDialog extends FocusDialog implements ActionListener, Documen
         mainPanel.add(textSearchPanel);
         mainPanel.addSpace(10);
 
+        addSizePanel(mainPanel);
+
         contentPane.add(mainPanel, BorderLayout.CENTER);
 
         contentPane.add(DialogToolkit.createOKCancelPanel(searchButton, cancelButton, getRootPane(), this), BorderLayout.SOUTH);
@@ -247,6 +269,32 @@ public class SearchDialog extends FocusDialog implements ActionListener, Documen
         checkInputs();
         showDialog();
     }
+
+
+    void addSizePanel(YBoxPanel mainPanel) {
+        YBoxPanel sizePanel = new YBoxPanel(10);
+        sizePanel.setBorder(BorderFactory.createTitledBorder(Translator.get("Size")));
+        XAlignedComponentPanel compPanel = new XAlignedComponentPanel(5);
+        GridBagConstraints gbc = ProportionalGridPanel.getDefaultGridBagConstraints();
+        gbc.weightx = 1.0;
+        JPanel sz = new ProportionalGridPanel(4, gbc);
+        sz.add(new JLabel(""));
+
+        cboSizeRel.setSelectedIndex(lastSizeRel.index);
+        sz.add(cboSizeRel);
+
+        tfSize.setText(lastSize == null ? "" : lastSize.toString());
+        sz.add(tfSize);
+
+        cboSizeUnit.setSelectedIndex(lastSizeUnit.index);
+        sz.add(cboSizeUnit);
+
+        compPanel.addRow("", sz, 10);
+        sizePanel.add(compPanel);
+        mainPanel.add(sizePanel);
+        mainPanel.addSpace(10);
+    }
+
 
     /**
      * Checks if search string or search location is empty (or white space), and enable/disable 'Add'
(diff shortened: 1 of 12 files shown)


mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -127,7 +145,8 @@ public SearchDialog(MainFrame mainFrame) {
         XAlignedComponentPanel compPanel = new XAlignedComponentPanel(5);
 
         String searchFor = PathUtils.removeLeadingSeparator(searchURL.getPath());
-        searchFilesField = new JTextField(searchFor);
+        if (searchFor.isEmpty()) searchFor = "*";
```
ahadas, line 148: formatting - please break this line
```java
if (searchFor.isEmpty())
    searchFor = "*";
```

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -274,7 +322,8 @@ public void actionPerformed(ActionEvent e) {
             return;
         }
         // otherwise, searchButton was pressed
-        updateValues();
+        if (!validateAndUpdateValues()) return;
```
ahadas, line 325: formatting - please break this line

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -248,6 +270,32 @@ private void cleanError() {
...
+        sz.add(tfSize);
+
+        cboSizeUnit.setSelectedIndex(lastSizeUnit.index);
+        sz.add(cboSizeUnit);
+
+        compPanel.addRow("", sz, 10);
+        sizePanel.add(compPanel);
+        mainPanel.add(sizePanel);
+        mainPanel.addSpace(10);
+    }
```
ahadas, line 296: yeah, this layout is not that nice :)
how about adding it as another row in the File Search panel with label `Size` and the component will be a panel with BorderLayout - the relation will be in the west, the size at the center and the unit at the east?

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -307,6 +360,24 @@ private void updateValues() {
         lastText = searchTextField.getText();
         lastTextCase = textCase.isSelected();
         lastTextRegex = textRegex.isSelected();
+
+        lastSizeRel = SizeRelation.VALUES[cboSizeRel.getSelectedIndex()];
+        lastSizeUnit = SizeUnit.VALUES[cboSizeUnit.getSelectedIndex()];
+
+        String s = tfSize.getText();
```
ahadas, line 367: how about naming it "size" rather than "s"?

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -348,6 +419,11 @@ private String getSearchQuery() {
             if (lastTextRegex)
                 properties.add(new Pair<>(SearchBuilder.TEXT_MATCH_REGEX, Boolean.TRUE.toString()));
         }
+        if (lastSize != null) {
+            properties.add(new Pair<>(SearchBuilder.SEARCH_SIZE_REL, lastSizeRel.name()));
+            properties.add(new Pair<>(SearchBuilder.SEARCH_SIZE, lastSize.toString()));
+            properties.add(new Pair<>(SearchBuilder.SEARCH_SIZE_UNIT, lastSizeUnit.toString() ));
+        }
```
ahadas, line 426: the problem I see with this approach is that we'll be using a single 'size' condition per-search
I can see us adding the possibility to search for files where, e.g., size > 10MB and size < 20MB and then we'll need to replace this representation. what do you think about encoding this in a way that support multiple size conditions?

mucommander-core/src/main/java/com/mucommander/search/SizeRelation.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+    },
+    gt(2,">"){
+        @Override
+        public boolean matches(long sz, long limit, SizeUnit unit) {
+            return sz > limit * unit.factor;
+        }
+    };
+
+    final int index;
```
ahadas, line 42: what do we need the `index` field for?

mucommander-core/src/main/java/com/mucommander/search/SizeUnit.java
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.mucommander.search;
+
+/**
+ * @author Gerolf Scherr
+ */
+public enum SizeUnit {
+    bytes(0,1), kB(1,1024), MB(2,1024*1024), GB(3,1024l*1024*1024);
```
ahadas, line 23: we should use the translatable units in `SizeFormat` (apparently they are different in some languages..)

mucommander-core/src/main/java/com/mucommander/ui/text/EnhancedTextField.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.swing.*;
+import javax.swing.text.JTextComponent;
+import java.awt.event.FocusEvent;
+import java.awt.event.FocusListener;
+
+
+/**
+ * @author Gerolf Scherr
+ */
+public class EnhancedTextField extends JTextField {
```
ahadas, line 28: I see that we always set `selectAllOnFocus` to `true` which makes sense ebcause otherwise it acts as `JTextField`, right? I wonder if we really want this field to always "select all" when gaining the focus or just when opening the dialog, if we do that only when opening the dialog then we can replace this class with a call to `selectAll`..

mucommander-translator/src/main/resources/dictionary.properties
```diff
@@ -484,6 +484,8 @@ search_dialog.case_sensitive = $[file_selection_dialog.case_sensitive]
 search_dialog.matches_regexp = $[file_selection_dialog.matches_regexp]
 search_dialog.wildcards = (* = any string, ? = any character, \\ = escape for literals: *?)
 search_dialog.search_text  = Containing text
+search_dialog.search_memonic=t
```
ahadas, line 487: this line should be placed elsewhere - here we should have a translatable strings whereas this looks like some kind of a configuration. we have configuration for keyboard shortcuts but they are associated with actions, I believe that this kind of shortcuts is not configurable anywhere - making it configurable should probably be out of the scope of this PR

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -55,14 +57,21 @@
 import com.mucommander.text.Translator;
 import com.mucommander.ui.action.ActionProperties;
 import com.mucommander.ui.action.impl.FindAction;
+import com.mucommander.ui.dialog.InformationDialog;
 import com.mucommander.ui.main.MainFrame;
+import com.mucommander.ui.text.*;
```
ahadas, line 62: formatting - we don't use wildcards in `import` statements

ahadas (commented):
thanks for posting this - it's really a useful functionality that is currently missing
I commented also on some formatting issues, note that we have a [formatter](https://github.com/mucommander/mucommander/blob/master/mucommander-code-format.xml) that should be usable via IntelliJ as well (not sure how to configure it there though)

mucommander-core/src/main/java/com/mucommander/search/SizeRelation.java
gerolfscherr: this is the index in the combobox / in the values(). we do need it to do JComboBox.setSelectedIndex(last.index)
 after thinking about it, this is the same functionality as Enum.ordinal(), i have changed it.

mucommander-core/src/main/java/com/mucommander/search/SizeUnit.java
gerolfscherr: I've done that.

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
gerolfscherr: I've played a little bit with the layouts (BorderLayout looks ugly,  GridLayout looks less ugly), now I have found one solution which is not as ugly as the other ones.

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
gerolfscherr: done

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
gerolfscherr: done

mucommander-translator/src/main/resources/dictionary.properties
gerolfscherr: I have removed it from the dictionary and set in in the code.

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
gerolfscherr: done

mucommander-core/src/main/java/com/mucommander/ui/text/EnhancedTextField.java
gerolfscherr: the select-all is the behaviour of TC (which I am quite used to), and I think the text field shouldn't behave differently when opening the dialog or jumping to the text field.

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
gerolfscherr: the easiest change would be to have SEARCH_SIZE_REL2 SIZE_2... for a second condition.

gerolfscherr:
@ahadas if there is anything I can do, please let me know!

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
ahadas: yeah, I guess we would have three options here:
1. search files with specific size (less common I guess)
2. search files with size lower than (or higher than) a certain value
3. search files with size that is between some min-value and max-value

so we won't  end up with more than two relations but naming the fields that way, with an index, will not be that great in the location field and the code for parsing it won't be that simple either

how about something like size=<:1:mb that indicates that we search for size that is lower than 1mb?
that way when handling the parsed properties for the search, we'll search for all occurrences of size properties and parse them the same way

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -248,6 +274,33 @@ private void cleanError() {
...
+        p.add(sp, BorderLayout.WEST);
+
+        JLabel l = compPanel.addRow(Translator.get("search_dialog.size", "Size"), p, 20);
+        l.setDisplayedMnemonic('s');
+        l.setLabelFor(cboSizeRel);
+    }
+
+
+    private String[] getSizeUnitDisplayStrings() {
+        return Arrays.stream(SizeUnit.VALUES).map(s->Translator.get("search_dialog.size_unit." + s.name() , s.name()))
```
ahadas, line 299: I believe the second argument to Translator#get is not needed since there's no placeholder in the translated string, right?

mucommander-core/src/main/java/com/mucommander/ui/text/SelectAllOnFocusListener.java
```diff
@@ -0,0 +1,49 @@
...
+    public static SelectAllOnFocusListener get() {
+        return InstanceHolder.me;
+    }
+
+    private SelectAllOnFocusListener() { }
+
+    @Override
+    public void focusGained(FocusEvent e) {
+        Object o = e.getSource();
+        if (o instanceof JTextComponent) ((JTextComponent)o).selectAll();
```
ahadas, line 43: please add line-break between the condition and the call to selectAll

mucommander-core/src/main/java/com/mucommander/ui/text/SelectAllOnFocusListener.java
```diff
@@ -0,0 +1,49 @@
...
+
+
+/**
+ * @author Gerolf Scherr
+ */
+public class SelectAllOnFocusListener implements FocusListener {
+
+    private final static class InstanceHolder {
+        final static SelectAllOnFocusListener me = new SelectAllOnFocusListener();
+    }
```
ahadas, line 32: what the benefit in having the `InstanceHolder` class? the static instance can rather be a member of the `SelectAllOnFocusListener` class

mucommander-translator/src/main/resources/dictionary.properties
```diff
@@ -484,6 +484,12 @@ search_dialog.case_sensitive = $[file_selection_dialog.case_sensitive]
 search_dialog.matches_regexp = $[file_selection_dialog.matches_regexp]
 search_dialog.wildcards = (* = any string, ? = any character, \\ = escape for literals: *?)
 search_dialog.search_text  = Containing text
+search_dialog.size=Size
```
ahadas, line 487: you can rather refer to existing translation here and in the units below with, e.g., $[size]

ahadas (commented):
I checked out the code and it's indeed nicer now!

mucommander-core/src/main/java/com/mucommander/ui/text/SelectAllOnFocusListener.java
gerolfscherr: this pattern avoids synchronization issues. the inner class will be initialized via the class loader, which runs synchronized. see:
https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom

ahadas: I see, I don't think the lazy evaluation is important in this case but if we try to minimize the memory consumption, it's also ok to create a new instance of the listener per-field - the search dialog is rarely used so objects would be allocated and garbage collected rarely in this context

mucommander-core/src/main/java/com/mucommander/ui/text/EnhancedTextField.java
ahadas: ok, fair enough

ahadas (approved):
@gerolfscherr  thanks for contributing this, I'll get this in since it's in good shape and let's polish it separately

PR #627, opened 2022-04-25
Added Desktop folder to Drive menu, use system icons for Drive button

pskowronek:
PR contains:
- Added Desktop folder to Drive menu (just under user home directory)
- Use system icons for current folder in Drive button - slightly ignoring preferences settings in this case. System icons (if available) are already being used for Drive menu, so I for the current folder they should be used, too. The preference settings as to whether to use system icons (never, apps, always) is more for table view, where people may find system icons to messy or making the UI too busy.


ahadas:
@pskowronek I pushed a patch that formats LocalFile and DrivePopupButton classes, can you please rebased this one to help me identify the changes?

pskowronek:
pls hold on, my rebase went really wrong :)

pskowronek:
pls use: https://github.com/mucommander/mucommander/pull/634

PR #628, opened 2022-04-26
Fix to let Reveal In XYZ work on search and parent folder #591

pskowronek:
Fix addressing bug #591 to Reveal In XYZ work when:
- in search results
- clicked on '..'

Tested on macOS **only**

Issue #591, opened 2022-03-13
Please make "Reveal in Finder" callable from search results


Andreas0602:
"Reveal in Finder" is not available in search results. There's no plausible reason for this restriction. Please make it available.

Andreas0602:
Thank you very much for your efforts, @pskowronek!

PR #631, opened 2022-04-28
Reveal in finder in search 2

pskowronek:
Fix addressing bug https://github.com/mucommander/mucommander/issues/591 to Reveal In XYZ work when:

    in search results
    clicked on '..'

It replaces https://github.com/mucommander/mucommander/pull/628

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ParentFolderAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ParentFolderAction.java
@@ -65,10 +65,7 @@ public abstract class ParentFolderAction extends MuAction implements ActivePanel
 
     private void toggleEnabledStateAdapter() {
         AbstractFile currentFolder = mainFrame.getActivePanel().getCurrentFolder();
-        if (currentFolder.getURL().getScheme().equals(SearchFile.SCHEMA))
-            setEnabled(false);
-        else
-            toggleEnabledState();
+        toggleEnabledState();
     }
 
     /////////////////////////////////
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ParentFolderAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ParentFolderAction.java
@@ -65,7 +65,11 @@ public abstract class ParentFolderAction extends MuAction implements ActivePanel
 
     private void toggleEnabledStateAdapter() {
         AbstractFile currentFolder = mainFrame.getActivePanel().getCurrentFolder();
-        toggleEnabledState();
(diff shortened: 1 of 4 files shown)


mucommander-core/src/main/java/com/mucommander/ui/action/impl/RevealInDesktopAction.java
```diff
@@ -55,21 +56,25 @@ public RevealInDesktopAction(MainFrame mainFrame, Map<String, Object> properties
...
     protected void toggleEnabledState() {
         AbstractFile currentFolder = mainFrame.getActivePanel().getCurrentFolder();
+        AbstractFile selectedFile = mainFrame.getActiveTable().getSelectedFile();
         setEnabled(currentFolder.getURL().getScheme().equals(LocalFile.SCHEMA)
                 && !currentFolder.isArchive()
-                && !currentFolder.hasAncestor(AbstractArchiveEntryFile.class));
+                && !currentFolder.hasAncestor(AbstractArchiveEntryFile.class)
+                || currentFolder.getURL().getScheme().equals(SearchFile.SCHEMA)
+                && selectedFile != null
+                && selectedFile.getURL().getScheme().equals(LocalFile.SCHEMA));
```
ahadas, line 65: we'd still want to disable it for archive entries that appear in the search results, right?

pskowronek: I've checked that and in archives 'reveal in XYZ' doesn't appear. The condition that parent is `Search` schema and the file must be `LocalFile` should be enough. Unless I don't see sth.

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ParentFolderAction.java
```diff
@@ -65,10 +65,7 @@ public ParentFolderAction(MainFrame mainFrame, Map<String, Object> properties) {
 
     private void toggleEnabledStateAdapter() {
         AbstractFile currentFolder = mainFrame.getActivePanel().getCurrentFolder();
-        if (currentFolder.getURL().getScheme().equals(SearchFile.SCHEMA))
-            setEnabled(false);
-        else
-            toggleEnabledState();
+        toggleEnabledState();
```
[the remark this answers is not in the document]
pskowronek: I will fix `RevealInDesktopAction` to use `ActiveTabAction` and revert changes in `ParentFolderAction` (though I will put braces to if/else over there).

mucommander-core/src/main/java/com/mucommander/ui/action/impl/RevealInDesktopAction.java
ahadas: right

ahadas (approved):
thanks!

PR #634, opened 2022-04-30
Added Desktop folder to Drive menu, use system icons for Drive menu

pskowronek:
PR contains:

- Added Desktop folder to Drive menu (just under user home directory)
- Use system icons for current folder in Drive button - slightly ignoring preferences settings in this case. System icons (if available) are already being used for Drive menu, so I for the current folder they should be used, too. The preference settings as to whether to use system icons (never, apps, always) is more for table view, where people may find system icons to messy or making the UI too busy.

### DIFF
--- a/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
+++ b/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
@@ -396,7 +396,7 @@ public class LocalFile extends ProtocolFile {
         try {
             AbstractFile desktop = homeFolder.getDirectChild("Desktop");
             if (desktop.exists() && desktop.isDirectory() && desktop.canRead()) {
-                volumesV.add(FileFactory.getFile(desktop.getAbsolutePath()));
+                volumesV.add(desktop);
             }
         } catch (IOException e) {
             LOGGER.debug("Thrown exception while getting Desktop folder", e);
--- a/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
+++ b/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
@@ -32,6 +32,7 @@ import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.StringTokenizer;
 import java.util.Vector;
@@ -306,6 +307,8 @@ public class LocalFile extends ProtocolFile {
         if (!(homeFolder == null || volumesV.contains(homeFolder)))
             volumesV.add(homeFolder);
 
+        addDesktopEntry(volumesV, homeFolder);
+
         AbstractFile volumes[] = new AbstractFile[volumesV.size()];
         volumesV.toArray(volumes);
 
@@ -377,6 +380,22 @@ public class LocalFile extends ProtocolFile {
             String warning =
                     "Error parsing" + (OsFamily.FREEBSD.isCurrent() ? "/sbin/mount -p output" : "/proc/mounts entries");
             LOGGER.warn(warning, e);
+        }
+    }
+
+    /**
+     * Adds Desktop to the given volumes entries if home directory is defined and Desktop folder
+     * is present and is writable (~/Desktop)
+     * @param volumesV the <code>Vector</code> to add mount points to
+     * @param homeFolder  a home folder, can be null
+     */
+    private static void addDesktopEntry(Vector<AbstractFile> volumesV, AbstractFile homeFolder) {
+        if (homeFolder == null) {
+            return;
+        }
+        File desktop = Paths.get(homeFolder.getAbsolutePath(), "Desktop").toFile();
+        if (desktop.exists() && desktop.canWrite() && desktop.isDirectory()) {
+            volumesV.add(FileFactory.getFile(desktop.getAbsolutePath()));
         }
     }
 
--- a/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
+++ b/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
@@ -393,9 +393,13 @@ public class LocalFile extends ProtocolFile {
         if (homeFolder == null) {
             return;
         }
-        File desktop = Paths.get(homeFolder.getAbsolutePath(), "Desktop").toFile();
-        if (desktop.exists() && desktop.canWrite() && desktop.isDirectory()) {
-            volumesV.add(FileFactory.getFile(desktop.getAbsolutePath()));
+        try {
+            AbstractFile desktop = homeFolder.getDirectChild("Desktop");
+            if (desktop.exists() && desktop.isDirectory() && desktop.canRead()) {
+                volumesV.add(FileFactory.getFile(desktop.getAbsolutePath()));
+            }
+        } catch (IOException e) {
+            LOGGER.debug("Thrown exception while getting Desktop folder", e);
         }
     }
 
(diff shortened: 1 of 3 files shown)


mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
```diff
@@ -380,6 +383,22 @@ private static void addMountEntries(Vector<AbstractFile> v) {
...
+     * @param homeFolder  a home folder, can be null
+     */
+    private static void addDesktopEntry(Vector<AbstractFile> volumesV, AbstractFile homeFolder) {
+        if (homeFolder == null) {
+            return;
+        }
+        File desktop = Paths.get(homeFolder.getAbsolutePath(), "Desktop").toFile();
+        if (desktop.exists() && desktop.canWrite() && desktop.isDirectory()) {
+            volumesV.add(FileFactory.getFile(desktop.getAbsolutePath()));
+        }
```
ahadas, line 399: how about "eating your own dog food" and use AbstractFile right from the beginning (rather than using Java's File at line 396)?

mucommander-core/src/main/java/com/mucommander/ui/main/DrivePopupButton.java
```diff
@@ -221,8 +221,12 @@ private void updateButton() {
 
             }
             setText(newLabel);
-            // Set the folder icon based on the current system icons policy
-            setIcon(FileIcons.getFileIcon(currentFolder));
+            { // local ctx for icon
```
ahadas, line 224: why do we need this block?

mucommander-core/src/main/java/com/mucommander/ui/main/DrivePopupButton.java
```diff
@@ -221,8 +221,12 @@ private void updateButton() {
...
+            // Set the folder icon slightly disobeying system icons policy for the current folder
+            Icon icon = FileIcons.hasProperSystemIcons() ?
+                    FileIcons.getSystemFileIcon(currentFolder) : FileIcons.getFileIcon(currentFolder);
```
ahadas, line 227: what's the reason for disobeying the system icons policy here?

ahadas (commented):
this PR introduces two changes: adding the Desktop folder, which looks good but I'd prefer us to use our own abstraction for files if possible, and showing the system file icons in the drive button regardless of the system icon policy which is debatable (why would we want the icon in the drive button to be different than the one within the table).

how about limiting this PR and the Desktop folder and discuss the icon in the drive button elsewhere?

mucommander-core/src/main/java/com/mucommander/ui/main/DrivePopupButton.java
pskowronek: I didn't want to create another method scope variable for icon. If you don't like it, then I can move `icon` outside of switch clause.

mucommander-core/src/main/java/com/mucommander/ui/main/DrivePopupButton.java
pskowronek: I was thinking that if somebody decided to have protocol icons then it work just like that (https://, gdrive:// etc).

pskowronek:
> this PR introduces two changes: adding the Desktop folder, which looks good but I'd prefer us to use our own abstraction for files if possible, and showing the system file icons in the drive button regardless of the system icon policy which is debatable (why would we want the icon in the drive button to be different than the one within the table).
> 
> how about limiting this PR and the Desktop folder and discuss the icon in the drive button elsewhere?

It is not so debatable - those systems icons are already being used for drive dropdown, so I see no reason not to use them for drive button. The reasoning behind letting ppl to decide as to whether use system icons or not, were more for table views when I admit using system icons maybe disturbing and for some may not look nice.

Screenshot how it looks after the change:
muIcons

mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
pskowronek: PR updated

pskowronek:
pls let me know if you still want to create a separate PR for those icons :)

mucommander-core/src/main/java/com/mucommander/ui/main/DrivePopupButton.java
pskowronek: PR updated

mucommander-core/src/main/java/com/mucommander/ui/main/DrivePopupButton.java
pskowronek: PR updated

mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/local/LocalFile.java
ahadas: I've made another minor change - using `desktop` that was retrieved from `homeFolder.getDirectChild("Desktop")` instead of retrieving it one again

ahadas (approved):
dropped the changes related to the icon in the drive button from this part so only those related to the 'Desktop' folder would remain, thanks!

ahadas:
@pskowronek about the system icons - I dropped them from this PR so it would focus only on adding the Desktop folder. did you mean there's another place we use system icons regardless of the system icons policy? if so, can you please point me where we do that?

pskowronek:
> @pskowronek about the system icons - I dropped them from this PR so it would focus only on adding the Desktop folder. did you mean there's another place we use system icons regardless of the system icons policy? if so, can you please point me where we do that?

In drive menu you can see those native icons. I dont have laptop with me now, i will try to provide exact classname and line numbere later today or tomorrow.

pskowronek:
> > @pskowronek about the system icons - I dropped them from this PR so it would focus only on adding the Desktop folder. did you mean there's another place we use system icons regardless of the system icons policy? if so, can you please point me where we do that?
> 
> In drive menu you can see those native icons. I dont have laptop with me now, i will try to provide exact classname and line numbere later today or tomorrow.

See here:
https://github.com/mucommander/mucommander/blob/74e5f9ba22aae9ee323e91c512630a3f040fe4d2/mucommander-core/src/main/java/com/mucommander/ui/main/DrivePopupButton.java#L469

ahadas:
@pskowronek ok, thanks, now I see what you meant. if you want the drive button to be consistent with that code then I'm ok with that but it should be similar to what this code does, i.e., setting volumes with system icons (not for every local file)

pskowronek:
> @pskowronek ok, thanks, now I see what you meant. if you want the drive button to be consistent with that code then I'm ok with that but it should be similar to what this code does, i.e., setting volumes with system icons (not every local file)

Can you try to run my changes and see how cool it looks like when you navigate from / (root) to your home, then to Desktop - the icon nicely changes from Macintosh HD native icon, thru home folder icon, and then special Desktop icon - everything provided by the system. Besides of being visually nice, those icons are very helpful to visually describe where we exactly are (I can provide more screenshots to convince you :)). Unless I misunderstood what you said (to only use the native icon for main folders of volumes?).

ahadas:
@pskowronek it's indeed nice, but it works the same way if you change the system icons policy to 'Always', right?

pskowronek:
@ahadas but then in file panels you also see them and many people may find it too chaotic (each file having a different system icon depending what application is associated with them). To kinda defend my stand for having native icons for drive button (besides already having those icons in drive popup menu) is that in muCommander app title the native icons are also already used to represent the current folder in the active panel:

obraz

So, the question is whether we would like to have the same native icon by default for drive menu, or extend icons policy to have another option: 'For applications and navigation only' (or rename 'For applications only').

ahadas:
> So, the question is whether we would like to have the same native icon by default for drive menu, or extend icons policy to have another option: 'For applications and navigation only' (or rename 'For applications only').

@pskowronek I'm afraid "navigation only" won't be so clear for users. I could think of adding another checkbox to the "Use system file icons" with "Show system icons in drive menu" so the first checkbox would apply to the table and there would be another setting for the drive menu. that said, I think it may be over engineering - I'm really not sure if users would actually want to have it displayed differently in the table and in the drive menu, I don't remember reports on that..

pskowronek:
So what now? Leave the topic or create an issue for it and collect feedback from users?

ahadas:
is it something that you find useful? I mean, regardless of what was the original purpose of adding the system icons policy, do you need this option of having system icons in the drive buttons but not in the tables? if so, feel free to post a PR that adds an additional configurable for this

pskowronek:
Yup, I find it useful. Will prepare PR for it.

PR #648, opened 2022-05-14
Native icons for drive button

pskowronek:
Added:
- Native icons for drive button (i.e. navigation)
- Additional preference to present native icons for applications and navigation (drive button)

To consider:
- is this preference name a good one?
- shouldn't the preference panel be reworked to have checkboxes (instead of dropdown) where those icons should be presented?


mucommander-preferences/src/main/java/com/mucommander/conf/SystemIconsPolicy.java
```diff
@@ -5,6 +5,10 @@ public enum SystemIconsPolicy {
     NEVER("never"),
     /** Use system file icons only for applications */
     APPLICATIONS_ONLY("applications"),
+
+    /** Use system file icons only for applications and navigation (drive button) */
+    APPLICATIONS_AND_NAVI_ONLY("applications_and_navigation"),
```
ahadas, line 10: this enumeration determines for which entities should the system icons be shown - everything / nothing / only-applications. It doesn't concern where because until now we didn't apply it anywhere besides the file table (right?). This way this value appear it seems to be read as "show system icons for applications within the file table or to everything in the navigation widgets". I believe we look for something different, another configuration that says "apply the selected system icons also to navigation widgets", right? in that case, I'd introduce another configuration value that appears as a checkbox for that

ahadas (commented):
@pskowronek thanks for posting this. yeah, I see why it was not easy to pick a good name and representation to this setting - commented inline with a suggestion for an alternative way

mucommander-preferences/src/main/java/com/mucommander/conf/SystemIconsPolicy.java
pskowronek: > this enumeration determines for which entities should the system icons be shown - everything / nothing / only-applications.

correct

> It doesn't concern where because until now we didn't apply it anywhere besides the file table (right?). 

nope, when you choose `Always`, then native icons are also being shown in navigation _aka_ drive button. The same for `Applications only` - when you dive into an application folder, then icon reflects the application (not very much use of it, rarely anybody dives into an application whatsoever).

> This way this value appear it seems to be read as "show system icons for applications within the file table or to everything in the navigation widgets". 

yes, and from UX perspective it is clear for me. From the code perspective (how this enum is being used) tho, it  more determines behaviors or rather what icon to show given the file/path.

> I believe we look for something different, another configuration that says "apply the selected system icons also to navigation widgets", right? in that case, I'd introduce another configuration value that appears as a checkbox for that

I took a look at the current preferences panel and honestly, I don't think it's gonna fit the bill from UX perspective to have separate/strayed checkbox appearing out of the moon to the right of the dropdown. 

If we wanted to have a better control and granularity, that dropdown should have options like: `Never`, `Always`, `Custom`. When `Custom` was used, then a bunch of the following checkboxes could be presented: `Applications`, `Navigation`, `Window title` etc. (side comment - this would be probably super clear and gave flexibility, but wouldn't be in line with current UX trends... but muC is all classic and that's why it is useful, right?)

ahadas: sure, adding more options would cover more scenarios but it's a bit over-engineering in this case

so I'm trying to recall how we got to this point:
1. we started with a PR that takes system icons for every location in the drive also when never/apps-only is selected
2. we ruled out saying "if you want system icons to be presented in the drive button, pick 'Always'" because it doesn't always look good at the file table
3. this PR introduces another value that is read as "apps-only in the file table or 'Always' in the drive button"

so how about another option that may satisfy everything - adding an option called 'folders-only" that would mean to take the system icons only for folders in the file table and the drive button. that way we won't have a "mess" in the file table, we keep the drive button and the file table persistent, and as you wrote above - it's unlikely for applications to be shown in the drive button so we don't lose much. what do you think?

pskowronek: > what do you think?

let me see how it's gonna like. I will back to you later on it :)

pskowronek: reworked, PR updated

pskowronek:
Rebased and modified the preference option (For folders only) and updated logic behind accordingly.

ahadas (approved):
thanks!

Issue #409, opened 2021-02-04
Run Command does not recognize shell alias


stef204:
Is it possible to have the `Run Command` recognize shell aliases? Using zsh on Linux.
I have `zsh -c` as `custom shell` under `Preferences|Misc`.

muCommander
Version: 0.9.6
Build date: 2021/01/02
Build number: b64d056

Java
Runtime version: 11.0.10
VM name: OpenJDK 64-Bit Server VM
VM version: 11.0.10+9
VM vendor: Oracle Corporation

OS
Name: Linux

ahadas:
related to #661

ahadas:
Version 1.1.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.1.0-1)

Issue #269, opened 2020-02-06
RFE: Option to close `Run command` dialog after command execution


mardukbp:
In version 0.9.3 it is possible to execute a command in the current directory. However, it is desirable to be able to execute a command on the selected file, like `java -jar`.

ahadas:
@mardukbp you mean in the `Run command` dialog, right? you'll need to provide the filename to the `java` command in your example, I don't see how we can avoid that. For this, one can use the copy path/name shortcut and then paste it in the dialog. Do you see a way to improve this?

mardukbp:
Thanks Arik. That helps. It would be nice to have a checkbox to choose to automatically close the run command window after the execution. 

Am 10. Februar 2020 14:08:21 MEZ schrieb Arik Hadas <notifications@github.com>:

ahadas:
@mardukbp yeah, that's something we can do :) thanks for the idea

ahadas:
related to #661

ahadas:
@mardukbp we integrated the operating system's terminal into the Run-command dialog so it would be possible to close the dialog after the command execution, not with a checkbox but rather with `<command>;exit` on Linux/macOS or `<command> & exit`

ahadas:
Version 1.1.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.1.0-1)

mardukbp:
@ahadas Perfect! Thanks!

PR #661, opened 2022-05-27
JediTerm as built in terminal in muCommander

pskowronek:
Added [JediTerm](https://github.com/JetBrains/jediterm) as built in terminal in muC so now muC behaves like the rest of old-school navigators like Norton Commander, FAR, Midnight Commander etc, i.e when Ctrl-O is pressed folder tables are being replaced with a new Terminal. Press Ctrl-O once again and Terminal is killed and folder tables are set visible again. This shortcut is already being used by some other action, so to check how it works you need to reconfigure in Preferences the shortcut for `OpenInOtherPanelAction` to use sth else.

Please take a look and tell me what you think.

Screenshot:
muC-term


Andreas0602:
From the description and the screenshot, this looks great. I would definitely appreciate to have this in muCommander.

pskowronek:
The build failed because apparently I've built JediTerm jars with a newer java version. Since this is a PoC and WiP, I can fix that later if decided to go with it.

ahadas:
Hi @pskowronek ,
Thanks for posting this. I didn't look at the code yet but as I also had a branch for integrating JediTerm, I can provide some early feedback based on the description and the screenshot:

I like how it was integrated in trolcommander, that shows the terminal below the file table. The main advantage compared to what I see in the screenshot above is that it's not "hidden" - users would be aware of it because it would just be there. Another advantage is that you allocate more space to the output this way. On the other hand, there should be some integration between the folder pane and the terminal - not only in terms of "pwd"  but also to be able to easily paste the path/name of the selected file - and for that it would be better to keep presenting the folder pane together with the terminal. One challenge I had when presenting the terminal below the folder pane is to present it with a "single line" (that keeps being updated according the the current folder and hidden when the current folder is not a local folder - although iirc sftp is also supported so maybe that can be an exception..), and so show more lines when the terminal gets the focus (iirc, trolcommander has the last part). 

Another part was the packaging with OSGI which was tricky but I don't remember why

pskowronek:
@ahadas why not backport the terminal from trolcommander?

Interesting thought about the terminal being shown below and being expandable when gaining the focus. Indeed this is interesting, however the bit about presenting the last line when unfocused seems tricky (obviously) - probably it would require to resize terminal panel to achieve the effect? This tho might be unpredictable sometimes i.e. to behave strangely (when `top' or `htop` command is executed, or anything else using `ncurses` or similar).  However, I think that the first iteration of such a feature should be rather simple, i.e. not overcomplicated too much - the next iteration can improve UX later (it might be even required to enhance JediTerm to do it more nicely, either via enhancement request or PR some mods to their project).

The question then is how such terminal should behave, for example:
1) should it be killed when hidden?
2) should it be a single instance across folder all panes (including tabs), if so
3) whether to try to change CWD when switching between left/right folders, tabs and changing the directory? (if possible to do so just like that in JediTerm)
4) what 'exit' command should do (and how to swiftly override it :))

I tend to follow Midnight Commander behavior - a single instance and trying to switch CWD if necessary and possible. As it comes to `exit` - it should kill the terminal, and one needs to Ctrl-O to respawn a new one.

ahadas:
> @ahadas why not backport the terminal from trolcommander?

Sure, we can get inspiration from trolcommander but it's not as simple as cherry-picking the relevant commits. Additionally, if you compare the way file-search is implemented in trolcommander and in mucommander, I think you'd find that it's better in mucommander with the results being presented within the file table instead of another window - so it's ok to get inspiration from it but there is some room for improvements sometimes

About what to present when the terminal is "collapsed" - as far as I remember, in total commander the terminal used to always show something like `<path> $ `, i.e., waiting to the next command within the current folder, when it didn't have focus. Yeah, it won't allow you to run something like 'top' and keep using file table at the same time but I don't think it's a common use of a terminal that is integrated within a file manager - arguably, it's more likely to run a command/script that you don't care about when you switch back the file table

> should it be killed when hidden?

In what approach? in the trolcommander's approach, it shouldn't allow you to switch focus away while there's something running in the terminal so if you switch to a different location that the terminal is not applicable for (e.g., dropbox), there's nothing to kill, just to hide the widget

> I tend to follow Midnight Commander behavior - a single instance and trying to switch CWD if necessary and possible.

+1

> As it comes to exit - it should kill the terminal, and one needs to Ctrl-O to respawn a new one.

Yeah, probably

pskowronek:
> > @ahadas why not backport the terminal from trolcommander?
> 
> Sure, we can get inspiration from trolcommander but it's not as simple as cherry-picking the relevant commits. Additionally, if you compare the way file-search is implemented in trolcommander and in mucommander, I think you'd find that it's better in mucommander with the results being presented within the file table instead of another window - so it's ok to get inspiration from it but there is some room for improvements sometimes

Agreed.

> About what to present when the terminal is "collapsed" - as far as I remember, in total commander the terminal used to always show something like `<path> $ `, i.e., waiting to the next command within the current folder, when it didn't have focus. Yeah, it won't allow you to run something like 'top' and keep using file table at the same time but I don't think it's a common use of a terminal that is integrated within a file manager - arguably, it's more likely to run a command/script that you don't care about when you switch back the file table

If it was the case, then a simple command line would be enough (even the existing one `run in terminal`). A full-blown terminal is useful for some advanced usage (`top` was just an example), but for example editing a file with `vim` etc is more probable. 

> > should it be killed when hidden?
> 
> In what approach? in the trolcommander's approach, it shouldn't allow you to switch focus away while there's something running in the terminal so if you switch to a different location that the terminal is not applicable for (e.g., dropbox), there's nothing to kill, just to hide the widget

In approach where terminal is being hidden. Having only last line shown - kinda hard to implement technically.
Midnight Commander let you return to table view even if a command is running - I like this approach.

> > I tend to follow Midnight Commander behavior - a single instance and trying to switch CWD if necessary and possible.
> 
> +1
+1  

> > As it comes to exit - it should kill the terminal, and one needs to Ctrl-O to respawn a new one.
> 
> Yeah, probably

Pls let me know if I should continue with current approach, i.e. fix jars, try to intercept `exit`? Like the first approach and later on some UX tweaks?

ahadas:
@pskowronek see how the terminal is integrated here:
"A Windows Guy in a Linux World: Midnight Commander on Linux" https://adamtheautomator.com/midnight-commander-linux/"

pskowronek:
> @pskowronek see how the terminal is integrated here: "A Windows Guy in a Linux World: Midnight Commander on Linux" https://adamtheautomator.com/midnight-commander-linux/"

yeah I know, I've been using MC for more than 20 years :) That's why I wanted similar behavior in muC to open full terminal when Ctrl-O is engaged. The one-liner you would like to have would be nice I agree, but I don't have idea how to achieve it with JediTerm easily. I would doing the terminal integration with small steps - first integrate it without one-liner, then try to add one-liner later on (probably with JediTerm enhancements and their help?).

ahadas:
> > @pskowronek see how the terminal is integrated here: "A Windows Guy in a Linux World: Midnight Commander on Linux" https://adamtheautomator.com/midnight-commander-linux/"
> 
> yeah I know, I've been using MC for more than 20 years :) That's why I wanted similar behavior in muC to open full terminal when Ctrl-O is engaged. The one-liner you would like to have would be nice I agree, but I don't have idea how to achieve it with JediTerm easily. I would doing the terminal integration with small steps - first integrate it without one-liner, then try to add one-liner later on (probably with JediTerm enhancements and their help?).

ack, let's try to make progress in steps and see where we get to when we'll have enough content for v1.1

pskowronek:
> > > @pskowronek see how the terminal is integrated here: "A Windows Guy in a Linux World: Midnight Commander on Linux" https://adamtheautomator.com/midnight-commander-linux/"
> > 
> > 
> > yeah I know, I've been using MC for more than 20 years :) That's why I wanted similar behavior in muC to open full terminal when Ctrl-O is engaged. The one-liner you would like to have would be nice I agree, but I don't have idea how to achieve it with JediTerm easily. I would doing the terminal integration with small steps - first integrate it without one-liner, then try to add one-liner later on (probably with JediTerm enhancements and their help?).
> 
> ack, let's try to make progress in steps and see where we get to when we'll have enough content for v1.1

OK, I will sync the branch with master and update / rebuild JediTerm jars with a correct java version.

mucommander-core/build.gradle
```diff
@@ -37,6 +56,17 @@ dependencies {
...
+        include 'com/intellij/**'
+        include 'com/pty4j/**'
+        include 'gnu/trove/**'
+        include 'jtermios/**'
+        include 'purejavacomm/**'
+        include 'kotlin/**'
+        include 'org/jetbrains/**'
+        include 'resources/com/pty4j/**'
+        include 'com/sun/**'
+        include 'org/apache/log4j/**'
```
ahadas, line 69: how much data does it add to mucommander-core jar?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Map;
+
+import javax.swing.KeyStroke;
+import javax.swing.SwingUtilities;
+
+/**
+ * This action shows built-in terminal (it mimics the behavior of Midnight Commander Ctrl-O command
+ * that originates back from Norton Commander).
+ */
+public class ShowTerminalAction extends ActiveTabAction {
```
ahadas, line 43: maybe it's time to add an osgi extension point for actions? that's something I thought about long time ago but didn't file a good justification but here I see that the functionality can be added as a module that register an action, what do you think?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
```diff
@@ -0,0 +1,140 @@
...
+
+    public static class Descriptor extends AbstractActionDescriptor {
+        public static final String ACTION_ID = "ShowTerminal";
+
+        public String getId() {
+            return ACTION_ID;
+        }
+
+        public ActionCategory getCategory() {
+            return ActionCategory.NAVIGATION;
```
ahadas, line 105: `navigation` is probably not the right category for this
'Run command' was placed in 'File operations' but that doesn't sound right either, maybe it should be 'misc'?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
```diff
@@ -0,0 +1,140 @@
...
+
+        public KeyStroke getDefaultKeyStroke() {
+            return KeyStroke.getKeyStroke(KeyEvent.VK_O, KeyEvent.CTRL_DOWN_MASK);
+        }
+
+        @Override
+        public String getLabel() {
+            return Translator.get(ActionProperties.getActionLabelKey(ShowTerminalAction.Descriptor.ACTION_ID),
+                    DesktopManager.canOpenInFileManager() ? DesktopManager.getFileManagerName()
+                            : Translator.get("file_manager"));
```
ahadas, line 120: copy-paste from RevealInDesktopAction?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/TerminalWindow.java
```diff
@@ -0,0 +1,75 @@
...
+        widget.setTtyConnector(createTtyConnector(currentFolder));
+        widget.start();
+        return widget;
+    }
+
+    private static @NotNull TtyConnector createTtyConnector(String currentFolder) {
+        try {
+            Map<String, String> envs = System.getenv();
+            String[] command;
+            if (UIUtil.isWindows) {
```
ahadas, line 61: we already have a per-OS module, how about moving this part to those (mucommander-os-*) modules?

mucommander-core/build.gradle
```diff
@@ -52,7 +82,7 @@ jar {
         'Implementation-Version': revision.substring(0, 7),
         'Build-Date': new Date().format('yyyyMMdd'),
         'Build-Url': "https://www.mucommander.com/version/nightly.xml",
-        'Import-Package': 'org.violetlib.aqua;resolution:=dynamic,com.formdev.flatlaf;resolution:=dynamic,com.apple.*;resolution:=dynamic,sun.security.action;resolution:=dynamic,*',
+        'Import-Package': 'org.violetlib.aqua;resolution:=dynamic,com.formdev.flatlaf;resolution:=dynamic,kotlin.internal.*;resolution:=dynamic,javax.jms;resolution:=dynamic,com.ibm.uvm.tools;resolution:=dynamic,sun.swing;resolution:=dynamic,org.apache.log4j;resolution:=dynamic,java.awt.*;resolution:=dynamic,sun.awt.*;resolution:=dynamic,kotlin.reflect.jvm.internal;resolution:=dynamic,com.sun.*;resolution:=dynamic,apple.awt.*;resolution:=dynamic,com.apple.*;resolution:=dynamic,sun.security.action;resolution:=dynamic,*',
```
ahadas, line 85: I tried to run the app with this change and mucommander-core failed with
```
org.osgi.framework.BundleException: Unable to resolve mucommander-core [102](R 102.0): missing requirement [mucommander-core [102](R 102.0)] osgi.ee; (&(osgi.ee=JavaSE)(version=14)) Unresolved requirements: [[mucommander-core [102](R 102.0)] osgi.ee; (&(osgi.ee=JavaSE)(version=14))]
```

I don't see yet where we added that dependency of Java 14 (I use Java 17), do you know?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
```diff
@@ -0,0 +1,140 @@
...
+                terminal = getTerminal();
+                mainFrame.getMainPanel().add(terminal, BorderLayout.CENTER);
+                terminal.revalidate();
+                SwingUtilities.invokeLater(() -> {
+                    terminal.requestFocusInWindow();
+                });
+
+                terminal.getTerminalPanel().addCustomKeyListener(new KeyAdapter() {
+                    public void keyPressed(KeyEvent keyEvent) {
+                        // TODO get the real configured key seq
```
ahadas, line 74: +1

ahadas (commented):
@pskowronek thanks again for posting this and sorry for the late feedback, this can be a really nice feature for the next release but there are some things we'll need to sort out, commented inside

pskowronek:
> @pskowronek thanks again for posting this and sorry for the late feedback, this can be a really nice feature for the next release but there are some things we'll need to sort out, commented inside

yup, I know. There are some other quirks that I have to fix first, for example I cannot set focus properly on panels after closing terminal. I will work on it though, unless someone has a clue how to fix it.

mucommander-core/build.gradle
```diff
@@ -24,8 +24,27 @@ dependencies {
...
+    comprise files('libs/annotations-20.1.0.jar')
+    comprise files('libs/jna-5.10.0.jar')
+    comprise files('libs/pty4j-0.12.7.jar')
+    comprise files('libs/terminal-2.69.jar')
+    comprise files('libs/eawtstub.jar')
+    comprise files('libs/jna-platform-5.10.0.jar')
+    comprise files('libs/purejavacomm-0.0.11.1.jar')
+    comprise files('libs/trove4j.jar')
+    comprise files('libs/jediterm-pty-2.69-SNAPSHOT.jar')
+    comprise files('libs/kotlin-stdlib-1.5.10.jar')
```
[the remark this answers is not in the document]
pskowronek: > at least some of those resources should appear on maven central and specifically jna is [already used](https://github.com/mucommander/mucommander/blob/1.0.1/mucommander-commons-file/build.gradle#L8-L9) so it would be better to get it once and have it shared between the modules that need it
> 
> as for the ones that are specific to the terminal, are they all built from https://github.com/JetBrains/jediterm or where did you get them from?

I took them from jediterm build folder. I will try to use maven for those libs which are there.

mucommander-core/build.gradle
pskowronek: I will check. The other question is whether `mucommander-core` is a good place to include terminal. Maybe a separate subsystem would be better? Another question is whether it works on Windows/Linux - I'm talking about changes made to osgi `Import-Package`.

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
pskowronek: Sounds good, would it mean that terminal and those import could be in a separate subsystem/module?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/TerminalWindow.java
pskowronek: hmm, I will take a look and comment later.

mucommander-core/build.gradle
pskowronek: I don't have a clue. What bothers me also (like I said in previous comments) is Windows/Linux and whether those imports will be fine for other OSes.

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
pskowronek: do you know whether such a trickery is already being done somewhere in the code?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
ahadas: yep, separate osgi bundle :)

mucommander-core/build.gradle
ahadas: those 'dynamic' ones shouldn't harm

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
ahadas: not from the top of my head, I'll need to check

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private void revertToTableView() {
+        LOGGER.info("Going to hide Terminal...");
+        if (terminal != null) {
+            mainFrame.getMainPanel().remove(terminal);
+            terminal.stop();
+        }
+        mainFrame.getSplitPane().setVisible(true);
+        SwingUtilities.invokeLater(() -> {
+            // TODO focus on active terminal....
```
pskowronek, line 136: FYI still it doesn't work as expected

mucommander-core/build.gradle
pskowronek: > how much data does it add to mucommander-core jar?

3MB -> 10MB

Those jediterm related jars take 7.5MB.

ahadas:
@pskowronek maybe it is mentioned somewhere and I missed it - we better clone the repositories that those jars were built from, can you please point me to those repos?

pskowronek:
> @pskowronek maybe it is mentioned somewhere and I missed it - we better clone the repositories that those jars were built from, can you please point me to those repos?

Do you mean git repo?
In `mu-commander/build.gradle` I've added a comment how those jars were created (basically, built from https://github.com/JetBrains/jediterm, commit: c21a2b27629dd3760c603edeb258f9a862e75d34)

And the artifactory they use for dependencies:
https://github.com/JetBrains/jediterm/blob/master/JediTerm/build.gradle#L24

ahadas:
@pskowronek I replaced most of the jars but still didn't identify where that requirement for Java 14 comes from - but I'm getting there..
I also managed to make it work with some tweaks on my side and it's nice! I noticed two issues while playing with that:
1. something about the focus doesn't work smoothly when closing the terminal - the file table doesn't always get the focus back it seems
2. when opening the terminal and closing it, it doesn't always start when trying to open it again - it seems that sometimes ctrl+o triggers something on the other panel.. didn't really investigate this though

I was thinking whether to integrate it as a replacement of the "console" for 'Run command', it will solve #409 and maybe we can some up with some solution for #269 as well. what do you think?

ahadas:
ok, that dependency came from the jediterm jar - but the weird thing is that when I compiled it locally, Java 14 was no longer needed

pskowronek:
@ahadas  

ad 1. & 2: I've tried to fix that with no luck. Clicking with mouse on file table fixes the problem, hitting Tab also (to some extend). Maybe we should "de-focus" file tables before hiding them or sth like that?

> I was thinking whether to integrate it as a replacement of the "console" for 'Run command',

Yes, I was thinking about it, too. I guess that 'console' was my initial trigger to search for something more advanced :)

pskowronek:
> @ahadas
> 
> ad 1. & 2: I've tried to fix that with no luck. Clicking with mouse on file table fixes the problem, hitting Tab also (to some extend). Maybe we should "de-focus" file tables before hiding them or sth like that?

I've just pushed a fix for both problems (it was sooo easy).

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
```diff
@@ -129,12 +142,12 @@ private void revertToTableView() {
...
             mainFrame.getMainPanel().remove(terminal);
-            terminal.stop();
         }
         mainFrame.getSplitPane().setVisible(true);
         SwingUtilities.invokeLater(() -> {
-            // TODO focus on active terminal....
-            mainFrame.getSplitPane().requestFocusInWindow();
+            // TODO - it seems like keyboard buffer is being tainted or focus is not complete/exact......
+            mainFrame.getActivePanel().requestFocusInWindow();
+            mainFrame.getActiveTable().requestFocusInWindow();
```
ahadas, line 149: I think the last line should be enough

ahadas:
> I've just pushed a fix for both problems (it was sooo easy).

yep, works well now!

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java
pskowronek: fix pushed, plus avoid changing CWD if the same as previous one. Looking for a way to change CWD w/o using `cd` (or at least try to find out how MC does it so they don't print `cd` in console)

ahadas:
@pskowronek I've made some changes to the dynamic imports and the terminal seems to work well with those changes on Linux (Fedora) and macOS. I didn't try with bundled JRE - we may need to modify it a bit to get javax.jms. can you give those [changes](https://github.com/ahadas/mucommander/commit/4446a3d1abe128235c663566b9b15398a51de03a) a try to your machine?

pskowronek:
> @pskowronek I've made few changes to the dynamic imports and the terminal seems to work well with those changes on Linux (Fedora) and macOS. I didn't try with bundled JRE though - we may need to modify it a bit to get javax.jms. can you give those [changes](https://github.com/ahadas/mucommander/commit/4446a3d1abe128235c663566b9b15398a51de03a) a try to your machine?

@ahadas I've cloned your repo with these changes and it work good (`./gradlew clean run`).

BTW, `./gradlew clean shadowJar` - isn't working? I mean, it finished in seconds with success apparently doing nothing.
Also tried `./gradlew dmg` but I got: 
```
Unable to wrap listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar
java.lang.IllegalArgumentException: Invalid syntax for version: 9999.0-empty-to-avoid-conflict-with-guava
```
I can try to run it on Windows 10, but can you tell what command exactly to execute to either build fat jar or exe?

ahadas:
> @ahadas I've cloned your repo with these changes and it work good (`./gradlew clean run`).

Great, thanks for checking that

> 
> BTW, `./gradlew clean shadowJar` - isn't working? I mean, it finished in seconds with success apparently doing nothing. Also tried `./gradlew dmg` but I got:
> 
> ```
> Unable to wrap listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar
> java.lang.IllegalArgumentException: Invalid syntax for version: 9999.0-empty-to-avoid-conflict-with-guava
> ```

Oh, ok this should be easy to solve by setting the Bundle-Version explicitly for guava
And yes, shadowJar is not really used anymore - there's still a reference for it within the launch4j task but I think it can be removed and shadowJar can be dropped

> 
> I can try to run it on Windows 10, but can you tell what command exactly to execute to either build fat jar or exe (`./gradlew createExe` does nothing)?

`createExe` produces an exe file in build/launch4j but let me try to fix that portable version

pskowronek:
> Oh, ok this should be easy to solve by setting the Bundle-Version explicitly for guava
> And yes, shadowJar is not really used anymore - there's still a reference for it within the launch4j task but I think it can be removed and shadowJar can be dropped

Would be nice to update readme.md about it (shadowJar is stated there).

I will try to run it on Windows.

ahadas:
> `createExe` produces an exe file in build/launch4j but let me try to fix that portable version

ok, it turned out to be more complex than I thought because the bundle is already set with version 31.1-jre which osgi can't work with but I've update [the terminal branch on my repo](https://github.com/ahadas/mucommander/tree/terminal) with a (temporary) fix

ahadas:
@pskowronek FYI, I removed `shadowJar` from the build scripts and updated the documentation
As for your last message above, it seems like you're missing the `jre` submodule.
Try: `git submodule init`, `git submodule update`
Then the scripts should bundle the JREs within the 'jre' submodule

pskowronek:
For DMG, now I got:
```
> Task :dmg FAILED
WARNING: Using incubator modules: jdk.incubator.jpackage
Error: Invalid Option: [--about-url]

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dmg'.
> Process 'command 'jpackage'' finished with non-zero exit value 1
```
To create EXE, is there anything I should run first before running `createExe`?

ahadas:
@pskowronek what version of JDK do you use? maybe the `--about-url` argument was added later - I use Java 17

createExe should be used as part of the portable version. If you want to create a Windows installer, `./gradlew msi` should be executed - but I've noticed some dependencies were missing on Windows, let me update my branch with those changes

ahadas:
@pskowronek updated my branch `terminal` with [another commit](https://github.com/ahadas/mucommander/commit/96530f519f69ae86c0a161bac2e063fb72f09aa2)

pskowronek:
> @pskowronek what version of JDK do you use? maybe the `--about-url` argument was added later - I use Java 17

jpackage from java 14 :) - `--help` doesn't show `--about-url` in that version.

> updated my branch terminal with [another commit](https://github.com/ahadas/mucommander/commit/96530f519f69ae86c0a161bac2e063fb72f09aa2) and created an [msi](https://github.com/mucommander/mucommander/releases/download/nightly/muCommander-1.1.0-test.msi) with those changes - can you give it a try?

Yes, I'll give it a try.

BTW, I think either readme.md or https://github.com/mucommander/mucommander/wiki/Packaging should be updated to describe the requirement for dev env in details (like `git submodule` etc)

pskowronek:
> @pskowronek updated my branch `terminal` with [another commit](https://github.com/ahadas/mucommander/commit/96530f519f69ae86c0a161bac2e063fb72f09aa2) and created an [msi](https://github.com/mucommander/mucommander/releases/download/nightly/muCommander-1.1.0-test.msi) with those changes - can you give it a try?

```
> Task :winAppImage FAILED
WARNING: Using incubator modules: jdk.incubator.jpackage
java.nio.file.NoSuchFileException: [...]/mucommander-ahadas/mucommander/build/muCommander.app/Contents/runtime/Contents/Home/lib/jli/libjli.dylib
```
/Of course I'm trying to cross-build msi on macOS/

ahadas:
@pskowronek I see that the changes I had to do for it to work on Windows are not included, do you have local changes that need to be posted or can I update this PR?

pskowronek:
> @pskowronek I see that the changes I had to do for it to work on Windows are not included, do you have local changes that need to be posted or can I update this PR?

I don't have any local changes.

pskowronek:
wow :)

pskowronek:
@ahadas thanks for merging. 

Do you have a hint how to get the configured shortcut?
https://github.com/mucommander/mucommander/blob/ab7aed6080bc046f75ebd12859a597a389aa6605/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java#L93

Andreas0602:
Thanks for your hard work, guys! Just one stupid question: How to close the terminal and return to the normal muCommander GUI?

pskowronek:
> Thanks for your hard work, guys! Just one stupid question: How to close the terminal and return to the normal muCommander GUI?

The same combination you used to enter, by default cmd-o / ctrl -o

Andreas0602:
> > Thanks for your hard work, guys! Just one stupid question: How to close the terminal and return to the normal muCommander GUI?
> 
> The same combination you used to enter, by default cmd-o / ctrl -o

Doesn't work for me. It doesn't close the terminal.

ahadas:
> @ahadas thanks for merging.

thank you :) good job on this!
 
> Do you have a hint how to get the configured shortcut?
> 
> https://github.com/mucommander/mucommander/blob/ab7aed6080bc046f75ebd12859a597a389aa6605/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ShowTerminalAction.java#L93

sure, see [this](https://github.com/mucommander/mucommander/commit/7bbd502be8108dee44cae621639e081dbb7d0444) (you can add it to ActionType)

ahadas:
> > > Thanks for your hard work, guys! Just one stupid question: How to close the terminal and return to the normal muCommander GUI?
> > 
> > 
> > The same combination you used to enter, by default cmd-o / ctrl -o
> 
> Doesn't work for me. It doesn't close the terminal.

Do you see a black terminal? we had an issue with that before but it doesn't happen to me with the changes that were merged on Windows/Linux/macOS so I wonder if you use the last version

Andreas0602:
> > > > Thanks for your hard work, guys! Just one stupid question: How to close the terminal and return to the normal muCommander GUI?
> > > 
> > > 
> > > The same combination you used to enter, by default cmd-o / ctrl -o
> > 
> > 
> > Doesn't work for me. It doesn't close the terminal.
> 
> Do you see a black terminal? we had an issue with that before but it doesn't happen to me with the changes that were merged on Windows/Linux/macOS so I wonder if you use the last version

Here are two screenshots:

Bildschirmfoto 2022-09-10 um 21 18 51

Bildschirmfoto 2022-09-10 um 21 18 24

ahadas:
@Andreas0602 ack, we were running the same version then. I see that you put the new command in the command-bar, did you also change its shortcut? switching back from the terminal to the file table should work with ctrl+o (regardless of the shortcut you defined to the action)

Andreas0602:
Yes, I changed the shortcut to F2. Pressing F2 opens the terminal. But pressing F2 again does not close it.

I have set up my Mac in a special way: I'm using a PC keyboard, and I'm running Karabiner Elements to "translate" ctrl+o to command+o. So, when I press ctrl+o, my Mac "sees" command+o. By the way, I've configured the same kind of translation for ctrl+f and ctrl+s. That's the reason why muCommander's internal editor was useless to me, because ctrl+f and ctrl+s did not work. Luckily, this has changed recently, when you changed the editor to use command+f and command+s on Mac. It would be good if you could make the same change for ctrl+o and the terminal. And it would be even better if you would use the same user-configured shortcut for closing the terminal as for opening it.

When I quit Karabiner Elements on my Mac, ctrl+o closes the terminal in muCommander.

pskowronek:
W/o Karabiner i.e. pure macOS Cmd-O it worked that way that the combination opened and closed the terminal. Haven't tried after merge tho. BTW, I need to work on getting the right key combination from configuration to close it properly (see https://github.com/mucommander/mucommander/pull/661#issuecomment-1242788768)

ahadas:
> W/o Karabiner i.e. pure macOS Cmd-O worked that way that the combination opened and closed the terminal. Haven't tried after the merge, tho. BTW, I need to work on getting the right key combination from configuration to close it properly (see [#661 (comment)](https://github.com/mucommander/mucommander/pull/661#issuecomment-1242788768))

@pskowronek I posted https://github.com/mucommander/mucommander/pull/726 for this, can you please take a look?

PR #676, opened 2022-06-22
File search with !subfolders & !archives & !symlinks won't call ls

ahadas:


Andreas0602:
By the way, the option "Search in subfolders" is not very clear. Even if you turn off that option, muCommander will still search in subfolders, as long as "Search depth" is greater than 1.
It looks like "Search in subfolders" has no effect, and only "Search depth" is relevant.

ahadas:
> By the way, the option "Search in subfolders" is not very clear. Even if you turn off that option, muCommander will still search in subfolders, as long as "Search depth" is greater than 1. It looks like "Search in subfolders" has no effect, and only "Search depth" is relevant.

Right, that has changed as well as part of this fix
If you don't select search-in-subfolders and don't select search-in-archives, the search depth wouldn't apply - only the root folder would be searched
If one of those (search-in-subfolders or search-in-archives) is selected, the search depth should apply

Andreas0602:
> > By the way, the option "Search in subfolders" is not very clear. Even if you turn off that option, muCommander will still search in subfolders, as long as "Search depth" is greater than 1. It looks like "Search in subfolders" has no effect, and only "Search depth" is relevant.
> 
> Right, that has changed as well as part of this fix If you don't select search-in-subfolders and don't select search-in-archives, the search depth wouldn't apply - only the root folder would be searched If at least one of those (search-in-subfolders or search-in-archives) is selected, the search depth should apply

Yes, I just installed and tested the latest nightly build, and I can confirm that it works as you've described above. Thank you.

PR #693, opened 2022-07-14
Pack local symbolic links to zip and tar archives

ahadas:


mucommander-archiver/src/main/java/com/mucommander/commons/file/archiver/ZipArchiver.java
```diff
@@ -94,4 +103,15 @@ public OutputStream createEntry(String entryPath, FileAttributes attributes) thr
     public void close() throws IOException {
         zos.close();
     }
+
+    @Override
+    public InputStream getContentStream(AbstractFile file) throws UnsupportedFileOperationException, IOException {
+        if (file.isSymlink()) {
```
ahadas, line 109: need to limit that to local files..

PR #712, opened 2022-08-10
fixed skopeo detection in registry plugin

danielerez:


mucommander-protocol-registry/src/main/java/com/mucommander/commons/file/protocol/registry/SkopeoCommandExecutor.java
```diff
@@ -18,8 +18,10 @@
 
 package com.mucommander.commons.file.protocol.registry;
 
+import com.fasterxml.jackson.databind.ser.std.StdKeySerializers.Default;
```
ahadas, line 21: not needed?

PR #726, opened 2022-09-10
Improve toggling to/from terminal view

ahadas:
Rename ShowTerminal action to ToggleTerminal since it is also responsible to switch back from the terminal view, and respect the configured shortcut or alternative-shortcut for switching back from the terminal view to the file table

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
@@ -21,10 +21,12 @@ import com.jediterm.terminal.ui.JediTermWidget;
 import com.mucommander.commons.file.AbstractFile;
 import com.mucommander.commons.file.protocol.local.LocalFile;
 import com.mucommander.core.desktop.DesktopManager;
+import com.mucommander.desktop.ActionType;
 import com.mucommander.text.Translator;
 import com.mucommander.ui.action.AbstractActionDescriptor;
 import com.mucommander.ui.action.ActionCategory;
 import com.mucommander.ui.action.ActionDescriptor;
+import com.mucommander.ui.action.ActionKeymap;
 import com.mucommander.ui.action.ActionProperties;
 import com.mucommander.ui.main.MainFrame;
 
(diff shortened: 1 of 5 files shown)


mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
```diff
@@ -90,9 +92,14 @@ public void performAction() {
 
                 terminal.getTerminalPanel().addCustomKeyListener(new KeyAdapter() {
                     public void keyPressed(KeyEvent keyEvent) {
-                        // TODO get the real configured key seq
-                        if (keyEvent.getKeyCode() == KeyEvent.VK_O &&
-                                (keyEvent.getModifiersEx() & KeyEvent.CTRL_DOWN_MASK) != 0) {
+                        KeyStroke pressedKeyStroke = KeyStroke.getKeyStrokeForEvent(keyEvent);
+                        KeyStroke keyStroke = ActionKeymap.getAccelerator(ActionType.ToggleTerminal.toString());
+                        if (pressedKeyStroke.equals(keyStroke)) {
```
pskowronek, line 97: I would make it a one condition or with 'else' instead

pskowronek (changes requested):
See my comment, everything else looks good

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
ahadas: ack, changed it to a single condition

Andreas0602:
Thank you for that change! I have now configured F10 to toggle the terminal, and it works well.

Some more comments:

1. The name of the action "ToggleTerminal.label" looks a bit ugly. "ToggleTerminal" without ".label" would look better.
2. When starting the terminal, it shows the text "The default interactive shell is now zsh. To update your account to use zsh, please run `chsh -s /bin/zsh`. For more details, please visit https://support.apple.com/kb/HT208050". Apple's Terminal.app does not do this. How to get rid of this message? Executing `chsh -s /bin/zsh` does not help.
3. When the terminal is open and I press F10, the terminal will be closed (that's fine) and I hear a short sound as if I'd pressed an invalid key (that's not okay -- there should be no such sound). Also, when I open the terminal again, there is the text "1~" in the command line. In the screenshot below, you can see the terminal after I've closed it three times and opened it a fourth time.
Bildschirmfoto 2022-09-11 um 09 54 02

pskowronek:
>     2. When starting the terminal, it shows the text "The default interactive shell is now zsh. To update your account to use zsh, please run `chsh -s /bin/zsh`. For more details, please visit https://support.apple.com/kb/HT208050". Apple's Terminal.app does not do this. How to get rid of this message? Executing `chsh -s /bin/zsh` does not help.

Do you have `zsh` as a default shell? 
Here I guess we would need to check what's the user's default shell is? https://github.com/mucommander/mucommander/blob/master/mucommander-core/src/main/java/com/mucommander/ui/action/impl/TerminalWindow.java#L74

Andreas0602:
> Do you have `zsh` as a default shell?

Yes.

pskowronek:
> Here I guess we would need to check what's the user's default shell is? https://github.com/mucommander/mucommander/blob/master/mucommander-core/src/main/java/com/mucommander/ui/action/impl/TerminalWindow.java#L74

https://stackoverflow.com/a/41553295/1715521 ?

Issue #729, opened 2022-09-12
Get rid of sound for 'invalid key pressed' when closing terminal


Andreas0602:
On macOS, when you have configured F10 to toggle the terminal and the terminal is open and you press F10, the terminal will be closed (that's fine) and you will hear a short sound as if you'd pressed an invalid key. That's not okay -- there should be no such sound.

Also, when you open the terminal again, there will be the text "1~" in the command line. That's not okay either.

<img width="828" alt="Bildschirmfoto 2022-09-12 um 08 36 10" src="https://user-images.githubusercontent.com/79788982/189588738-1ce4969d-2089-4a7d-986b-42b0b6cbe68a.png">

ahadas:
@pskowronek hi, could you look at this? also happened to me but it's not always reproducible

pskowronek:
The PR contains:
- fix for this issue
  - the custom key listener didn't mark the key event as `consumed`
  - however, JediTerm doesn't check if the event was consumed (I've created a bug and PR for it: https://github.com/JetBrains/jediterm/issues/251), and in the meantime, I've committed a workaround to muC to fix that
- some code clean up
- found that Cmd-O/Ctrl-O is also used to 'open in other tab' action - changed that default key configuration to Cmd-Alt-Shift-O/Ctrl-Alt-Shift-O so Terminal could be opened with well known/used combination Cmd-O/Ctrl-O.

@ahadas, BTW, I'm having problems importing muC to Idea - strangely, the library produced by `jetbrains-jediterm` subsystem is not recognized for core subsystem :/

ahadas:
@pskowronek thanks! all but the changing the default key shortcut for 'open in other tab' looks good to me - commented on the PR
about Idea - I have no idea :) I don't use IntelliJ, maybe it uses some generated configuration from the gradle one that needs to be refreshed? things like apache-bzip2 and sevenzipjbindings are recognized well?

PR #732, opened 2022-09-14
Fixes to built-in JediTerm regarding hot key handling #729

pskowronek:
Fix for https://github.com/mucommander/mucommander/issues/729

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
@@ -141,6 +141,10 @@ public class ToggleTerminalAction extends ActiveTabAction {
                 KeyStroke alternateAccelerator = ActionKeymap.getAlternateAccelerator(
                         ActionType.ToggleTerminal.toString());
                 if (pressedKeyStroke.equals(accelerator) || pressedKeyStroke.equals(alternateAccelerator)) {
+                    keyEvent.consume();
+                    // FIXME workaround, since JediTerm doesn't check #isConsumed status
+                    // https://github.com/JetBrains/jediterm/issues/251
+                    keyEvent.setKeyCode(-1);
                     revertToTableView();
                     setVisible(false);
                 }


mucommander-os-api/src/main/java/com/mucommander/desktop/ActionShortcuts.java
```diff
@@ -138,11 +138,11 @@ public KeyStroke getDefaultKeystroke(ActionType actionId) {
         case OpenInNewTab:
             return KeyStroke.getKeyStroke(KeyEvent.VK_ENTER, KeyEvent.CTRL_DOWN_MASK);
         case OpenInOtherPanel:
-            return KeyStroke.getKeyStroke(KeyEvent.VK_O, KeyEvent.CTRL_DOWN_MASK);
+            return KeyStroke.getKeyStroke(KeyEvent.VK_O, KeyEvent.SHIFT_DOWN_MASK | KeyEvent.CTRL_DOWN_MASK | KeyEvent.ALT_DOWN_MASK);
```
ahadas, line 141: I prefer not to change existing shortcuts as much as possible since users may have got used to the current configuration. in the long term, I think we would drop the terminal action in favor of a terminal widget below the file table and then we could use CMD+R / CTRL+R for switching the focus to that component (and getting rid of the run-command dialog)
we can use this shortcut (SHIFT+CTRL+ALT+O) to the terminal action in the meantime

pskowronek: Since all the commanders I know (that have terminal command) use Ctrl/CMD+O, I would suggest to follow this pattern.

ahadas: @pskowronek I see your point but it will place us in a not-that-great place if we get complains about "suddenly" changing the shortcut of 'open in other panel'. I also don't think it matters that much - as you've noticed, @Andreas0602 that I believe is the first user that uses this feature has already configured a different shortcut (F10) for it, and in the long term I'd like to drop that action when we'll have an integrated widget..

ahadas (approved):
@pskowronek I separated out the refactoring/cleanup part to #736 and dropped the changes to the shortcut of 'open in other panel' from this PR (which can be further discussed separately) so the fix for the issue will be isolated - good job!

Issue #728, opened 2022-09-12
Get rid of message "The default interactive shell is now zsh..." when starting terminal


Andreas0602:
On macOS, when your default shell is zsh and you start the terminal, the message "The default interactive shell is now zsh..." appears. This should not happen. Apple's Terminal.app does not do this either.

The reason for this problem is probably here: https://github.com/mucommander/mucommander/blob/master/mucommander-core/src/main/java/com/mucommander/ui/action/impl/TerminalWindow.java#L74

<img width="893" alt="Bildschirmfoto 2022-09-12 um 08 27 03" src="https://user-images.githubusercontent.com/79788982/189587302-391fcef9-e7b2-41e1-84ab-f9a332b538a9.png">

pskowronek:
The fix is to get a default shell for a current user under macOS (only), idea taken from: https://stackoverflow.com/a/41553295/1715521

For other OSes it uses a hardcoded shells (cmd, /bin/bash).

Andreas0602:
Okay, from a formal point of view, this change addresses my original request to get rid of the message. But it doesn't really solve the underlying problem, it just hides it. The real solution should be to start the user's default login shell. On my Mac, both `dscl . -read ~/ UserShell` and `echo $SHELL` return the correct result, which is `UserShell: /bin/zsh` or `/bin/zsh` respectively in my case. @ahadas wrote [here](https://github.com/mucommander/mucommander/pull/733#issuecomment-1250040955) that `dscl . -read ~/ UserShell` doesn't work for him, so what about reading `$SHELL`?

ahadas:
@Andreas0602 I believe that using `dscl . -read ~/ UserShell` is the right way to go, it returned /bin/bash because that's the terminal that is configured on my machine. the idea behind #733 is good, just need to align it with the other configuration for the terminal, and maybe also make use of Unix4j to get the output from this command to save a few lines. feel free to chime in the thread there

PR #733, opened 2022-09-15
Get a default shell for current user under macOS #728

pskowronek:
PR for https://github.com/mucommander/mucommander/issues/728


ahadas:
@pskowronek I tried the posted change and it didn't remove the deprecation warning for me - the configured shell I got from `dscl . -read ~/ UserShell` was /bin/bash and so nothing really changed. I posted another PR that sets the environment variable BASH_SILENCE_DEPRECATION_WARNING to silent that warning but I like the direction of this PR nevertheless - how about trying to use DesktopAdapter#getDefaultShell so if the user configures a custom shell we'll take it?

pskowronek:
Regarding the warning - if you have /bin/bash by default, and no BASH_SILENCE_DEPRECATION_WARNING=1 in profile then you shouldn't see the warning. If you open Terminal then you have this warning or not? I wouldn't set BASH_SILENCE_DEPRECATION_WARNING explicitly in muC, tho. The terminal should start as if it was started by Terminal app etc.

I saw DesktopAdapter#getDefaultShell and it is more for command execution and not for interactive shell.

I was wondering if shell shouldn't be configurable (via preferences), to let user set the shell to their liking.

ahadas:
> Regarding the warning - if you have /bin/bash by default, and BASH_SILENCE_DEPRECATION_WARNING=1 in profile then you shouldn't see the warning. If you open Terminal then you have this warning or not?

yes, I see that warning when opening the terminal since I didn't set the env var BASH_SILENCE_DEPRECATION_WARNING

> I wouldn't set BASH_SILENCE_DEPRECATION_WARNING explicitly in muC, tho. The terminal should start as if it was started by Terminal app etc.

generally, I'd agree - but in the long term I would like us to replace the separate terminal window with a widget. I remember seeing it with previous file managers I used in the past - in my opinion it's a better user experience then switching between the windows, and then I wouldn't want to have those three lines that take space

> I saw DesktopAdapter#getDefaultShell and it is more for command execution and not for interactive shell.

yes, it's used by the run-command dialog which I hope to replace with jediterm and later, to drop that dialog when we have that integrated widget

> I was wondering if shell shouldn't be configurable (via preferences), to let user set the shell to their liking.

+1
and with the changes described above we can have a single configuration for the shell that will be used in both the terminal window and run-command dialog

ahadas:
On a second thought, it could have been better to clear the screen completely when initiating the terminal..

pskowronek:
> @pskowronek, how about polishing #733? is there something else you work on for v1.1.0?

@ahadas  - so, what's the decision? What should be done here for v1.1.0?

ahadas:
@pskowronek how about:
1. retrieve the default as you do here (the one that appears in Preferences -> Misc -> Run command
2. use that, or the custom shell that the user configures there, for jediterm

pskowronek:
> @pskowronek how about:
> 
>     1. retrieve the default as you do here (the one that appears in Preferences -> Misc -> Run command

Two questions here:
1) But this pref:
obraz
is for running commands (and where you have additional params to launch that command).
In case of terminal we should what? Remove those params and run just the command?

2) Do you mean to use my code to work out macOS' user default shell (like bash or zsh) and use it in pref? But with additional parameters to run the command? We would need to have a list of default params to run a command per known shell (bash, zsh, sh, and whatever else is common out there).

> 
>     2. use that, or the custom shell that the user configures there, in jediterm

But remove parameters? (I'm thinking about `Open Terminal` usage, not run command).

ahadas:
> 1. But this pref:
>    obraz
>    is for running commands (and where you have additional params to launch that command).
>    In case of terminal we should what? Remove those params and run just the command?

Yes, it makes sense to remove/replace those params in my opinion since we no longer just execute commands but rather initialize a terminal from which the user executes commands

> 2. Do you mean to use my code to work out macOS' user default shell (like bash or zsh) and use it in pref?

Yep

> But with additional parameters to run the command? We would need to have a list of default params to run a command per known shell (bash, zsh, sh, and whatever else is common out there).

We already have default params, the `--login`, right? so yeah, to list that and let users override the command and and its parameters if they wish to

> > ```
> > 2. use that, or the custom shell that the user configures there, in jediterm
> > ```
> 
> But remove parameters? (I'm thinking about `Open Terminal` usage, not run command).

I don't think we need to differentiate the terminal view and 'run command' here - is there a reason not to do the same thing for both?

pskowronek:
What about this part?
obraz

I'm looking at the usage of it (`Shell`, `ShellEncodingListener` etc) and encoding autodetection is being done by analyzing the output of the process - doesn't seem to fit JediTerm I guess. Any suggestions?

pskowronek:
FYI - The last PR update was just a rebase to sync with master (and it didn't go smooth :/)

pskowronek:
> What about this part? obraz
> 
> I'm looking at the usage of it (`Shell`, `ShellEncodingListener` etc) and encoding autodetection is being done by analyzing the output of the process - doesn't seem to fit JediTerm I guess. Any suggestions?

So, if we are replacing old run in terminal with JediTerm, I think the whole package `com.mucommander.shell` (Shell, ShellHistoryWriter/Reader/Manager etc) should be removed then. If we want to have autodetect still, then probably things stated in #750 should be used to workout proper settings for LC_ALL/LANG if those env vars are not present.

ahadas:
> What about this part? obraz
> 
> I'm looking at the usage of it (`Shell`, `ShellEncodingListener` etc) and encoding autodetection is being done by analyzing the output of the process - doesn't seem to fit JediTerm I guess. Any suggestions?

I'd drop the auto-detection then

pskowronek:
@ahadas PR updated:
- implemented suggested approach for using shell from prefs
- includes a fix for #750 
- removed files from `com.mucommander.shell` package and any of their references

ahadas:
yeah, something with the rebase looks wrong, I'll have a look

ahadas (approved):
@pskowronek I've made some cleanup and minor changes - could you please check that this version is alright?

pskowronek:
> @pskowronek I've made some cleanup and minor changes - could you please check that this version is alright?

@ahadas looks good!

PR #737, opened 2022-09-17
Change default shortcut of ToggleTerminal

ahadas:
The current shotcut (ctrl+O / meta+O) conflicts with that of 'Open in other panel' action and therefore is changed.


ahadas:
@pskowronek we need to resolve the conflict we currently have between the actions and as I wrote on #732, I don't want to change the shortcut of 'Open in other panel', do you have a better idea than the posted change?

pskowronek:
F12 on all platforms?

PR #739, opened 2022-09-17
Integrate native terminal in run-command dialog

ahadas:


ahadas:
@pskowronek it's still a work in progress but can you please have a look?

pskowronek:
> @pskowronek it's still a work in progress but can you please have a look?

@ahadas, so far it looks good. 

btw, I've just spotted - I see resize problem with toolbar and file panels when you resize the whole main window. Apparently it was introduce somewhere between 646ef13c6 and the current master (not in this PR).

ahadas:
> > @pskowronek it's still a work in progress but can you please have a look?
> 
> @ahadas, so far it looks good.

ack, thanks :)

> 
> btw, I've just spotted - I see resize problem with toolbar and file panels when you resize the whole main window. Apparently it was introduced somewhere between [646ef13](https://github.com/mucommander/mucommander/commit/646ef13c62b5142b2a00246f2982154921cca5da) (here it still works well) and the current master (not in this PR).

I don't see a problem on my machine, could you share a screenshot that demonstrates the problem maybe?

pskowronek:
> I don't see a problem on my machine, could you share a screenshot that demonstrates the problem maybe?

hold on....

pskowronek:
The problem is in master, steps to reproduce:
- resize main window, everything inside resize properly
- F12 (open terminal)
- F12 (close terminal)
- now things inside don't resize properly:

obraz

ahadas:
@pskowronek yeah, ok, this happens.. :) you're already looking into it, right?

pskowronek:
> @pskowronek yeah, ok, this happens.. :) you're already looking into it, right?

Yup

pskowronek:
@ahadas fix/workaround PR here: https://github.com/mucommander/mucommander/pull/742

Andreas0602:
When you open the terminal window by means of "run command", you cannot switch back to the main window of muCommander and perform some other tasks there. The only way is to close the terminal. This is inconvenient. Could this be changed? Or is this by design? If so, this would be a disadvantage compared to toggling the terminal window with F12.

Another point: Is there a special reason why the terminal opened by "run command" is black on white, while the terminal opened by F12 is white on black? It seems to be inconsistent.

ahadas:
> When you open the terminal window by means of "run command", you cannot switch back to the main window of muCommander and perform some other tasks there. The only way is to close the terminal. This is inconvenient. Could this be changed? Or is this by design? If so, this would be a disadvantage compared to toggling the terminal window with F12.

Yeah, the 'run command' dialog was designed to be a modal window. I don't want to put efforts in changing it now - the idea is rather to integrate the terminal below the file table as a replacement of both that 'run command' dialog and the new terminal window
 
> Another point: Is there a special reason why the terminal opened by "run command" is black on white, while the terminal opened by F12 is white on black? It seems to be inconsistent.

The reason for it is that it's the default theme that comes with mucommander - you can change it by going to Preferences -> Appearance -> Themes -> Edit -> Shell and choose the colors you'd like to have but note that the selection colors don't propagate to the new terminal..

Andreas0602:
So, as I understand it, you're planning to remove "run command". That would be okay from my point of view, because the modal window of "run command" is a significant disadvantage, so there's no point for me in using it.

Furthermore, as I understand it, you're planning to have the new terminal window below the file table. That means, it would only be one single line, right? That would also be okay from my point of view, **as long as it's still possible to make it bigger so that will occupy the whole muCommander window (except the toolbar at the top), like it is now when you press F12.** Please do not remove that capability.

It would be good if the settings made under Preferences -> Appearance -> Themes -> Edit -> Shell would be propagated to the new terminal.

ahadas:
> So, as I understand it, you're planning to remove "run command". That would be okay from my point of view, because the modal window of "run command" is a significant disadvantage, so there's no point for me in using it.

Yes, that's the plan - to drop the 'run command' dialog when the terminal is better integrated

> Furthermore, as I understand it, you're planning to have the new terminal window below the file table. That means, it would only be one single line, right? That would also be okay from my point of view, **as long as it's still possible to make it bigger so that will occupy the whole muCommander window (except the toolbar at the top), like it is now when you press F12.** Please do not remove that capability.

Well, there should be a single line when the folder table got the focus but when the focus switches to the terminal it can expand. I wouldn't set the default size as you wrote about (to occupy the whole window) but if you resize the terminal window, we should remember the size that was being set for the next time the terminal gets the focus

> It would be good if the settings made under Preferences -> Appearance -> Themes -> Edit -> Shell would be propagated to the new terminal.

+1
We didn't give much attention to the shell for a very long time but once we have an integrated shell/terminal, we can also reconsider the default colors

PR #742, opened 2022-09-18
Fix/workaround for SplitPane not auto-resizing after closing Terminal

pskowronek:
Fix/workaround for https://github.com/mucommander/mucommander/pull/739#issuecomment-1250332115

Please see a comment in the code and let me know what you think.


ahadas (approved):
thanks @pskowronek
yeah, there's probably a more elegant solution than that but I would rather spend the time on replacing this separated window for the terminal with something that is better integrated in the main window as we discussed before

PR #744, opened 2022-09-20
Faster start-up when Bonjour is enabled

pskowronek:
Faster start-up when Bonjour is enabled. I've spotted that init'n of Bonjour services consume like 3-5s (sometimes more if there's network problem) on start-up. I simply moved the initialization part into background, and while Bonjour is initializing, menu will display 'Bonjour initalizing...'.

Disclaimer: I haven't made substantial refactor.


mucommander-translator/src/main/resources/dictionary.properties
```diff
@@ -546,6 +546,7 @@ vsphere_connections_dialog.guest_password = Guest Password
 bonjour.bonjour_services = Bonjour services
 bonjour.no_service_discovered = No service discovered
 bonjour.bonjour_disabled = Bonjour disabled
+bonjour.bonjour_initializing = Bonjour initializing...
```
ahadas, line 549: we use to add the `...` postfix to operations that trigger opening a dialog/window and it's not the case here. is it at least disabled?

mucommander-core/src/main/java/com/mucommander/bonjour/BonjourMenu.java
```diff
@@ -90,8 +90,9 @@ public void menuSelected(MenuEvent menuEvent) {
                 // Inform that no service have been discovered
                 add(new JMenuItem(Translator.get("bonjour.no_service_discovered"))).setEnabled(false);
             }
-        }
-        else {
+        } else if (BonjourDirectory.isStarting()) {
+            add(new JMenuItem(Translator.get("bonjour.bonjour_initializing"))).setEnabled(false);
```
ahadas, line 94: I think that I never saw an entry in popup menu with indication of something that is being initializing/loading before, I guess it's not common because those menus are not refreshed when the data is available.. I think it would actually be better to wait here until initialization is completed - it would almost never happen in practice.. (that the menu is shown while Bonjour is being initialized)

mucommander-core/src/main/java/com/mucommander/Application.java
```diff
@@ -399,8 +400,11 @@ private void run() {
             com.mucommander.ui.theme.ThemeManager.loadCurrentTheme();
 
             // Starts Bonjour services discovery (only if enabled in prefs)
-            printStartupMessage("Starting Bonjour services discovery...");
-            com.mucommander.bonjour.BonjourDirectory.setActive(MuConfigurations.getPreferences().getVariable(MuPreference.ENABLE_BONJOUR_DISCOVERY, MuPreferences.DEFAULT_ENABLE_BONJOUR_DISCOVERY));
+            boolean bonjourEnabled = MuConfigurations.getPreferences().getVariable(MuPreference.ENABLE_BONJOUR_DISCOVERY, MuPreferences.DEFAULT_ENABLE_BONJOUR_DISCOVERY);
+            if (bonjourEnabled) {
+                printStartupMessage("Starting Bonjour services discovery...");
+            }
+            CompletableFuture.runAsync(() -> com.mucommander.bonjour.BonjourDirectory.setActive(bonjourEnabled));
```
ahadas, line 407: how about moving the async part to the BonjourDirectory class (i.e., making setActive become an async operation) - it might be useful for the Preferences dialog as well

mucommander-translator/src/main/resources/dictionary.properties
pskowronek: yes it is, it works the same as for `bonjour.bonjour_disabled`. Should I remove "..." ?

mucommander-core/src/main/java/com/mucommander/bonjour/BonjourMenu.java
pskowronek: I see it when I open muC and navigate to Bonjour - it initializes in the background for a couple of seconds. And as I stated in PR, for slow/strange networks it make take even longer.

mucommander-core/src/main/java/com/mucommander/Application.java
pskowronek: will take a look

mucommander-translator/src/main/resources/dictionary.properties
ahadas: yes please

mucommander-core/src/main/java/com/mucommander/bonjour/BonjourMenu.java
ahadas: that's ok not to block on it during startup but I'd say that when opening the popup menu, it's ok to wait for it to complete rather than indicating that it's being initialized

pskowronek: I would rather want to see that Bonjour is still fighting to gather services instead either having whole Bonjour item not present, or the whole popup menu blocked.

pskowronek:
@ahadas PR updated

mucommander-core/src/main/java/com/mucommander/bonjour/BonjourMenu.java
ahadas: ok, not that critical - it (opening the popup menu while Bonjour is initializing) should rarely happen

package/readme.txt
```diff
@@ -44,6 +44,7 @@ Improvements:
 - Keyboard shortcuts on macOS better align with the default keyboard shortcuts in Safari.
 - The 'About' dialog indicates when the portable version is used.
 - Commands that are executed from the 'Run command' dialog are now typed within a terminal that recognizes shell aliases.
+- Faster launch when Bonjour services enabled (initialization done asynchronously)
```
ahadas, line 47: how about:
```suggestion
- Bonjour services are initialized asynchronously in order to accelerate startup time.
```

pskowronek:
PR updated

ahadas (approved):
@pskowronek thanks! please rebase again - I've updated the readme file which lead to another conflict..

PR #754, opened 2022-09-25
Quick Search to work with Polish diacritics

pskowronek:
Quick Search to work with Polish diacritics typed using the most popular in PL keyboard layout: Polish (Programmer) (see [here](https://commons.wikimedia.org/wiki/File:Polish_programmer%27s_layout.PNG)).

This also applies also to any other keyboard layouts that use ALT+letter to get diacritics.

Also, sorting in FileTable (in panels) doesn't properly handle diacritics (FileComaparator#compare -> criterion == NAME_CRITERION) - diacritics don't sort properly (java.text.Collator should be used for current Locale). But this is sth for a separate issue.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/quicksearch/QuickSearch.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/quicksearch/QuickSearch.java
@@ -17,13 +17,13 @@
 
 package com.mucommander.ui.quicksearch;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.awt.event.KeyAdapter;
 import java.awt.event.KeyEvent;
 
 import javax.swing.JComponent;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * This class contains 'quick search' common functionality - selection of rows that match
@@ -120,7 +120,8 @@ public abstract class QuickSearch<T> extends KeyAdapter implements Runnable {
      * <code>false</code> in any of the following cases:
      *
      * <ul>
-     *   <li>has any of the Alt, Ctrl or Meta modifier keys down (Shift is OK)</li>
+     *   <li>has any of the Ctrl or Meta modifier keys down (Shift is OK)</li>
+     *   <li>has Alt but it is not a letter (in some languages diacritics are typed with ALT+letter)</li>
      *   <li>is an ASCII control character (<32 or ==127)</li>
      *   <li>is not a valid Unicode character</li>
      * </ul>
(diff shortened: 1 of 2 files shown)


mucommander-core/src/main/java/com/mucommander/ui/quicksearch/QuickSearch.java
```diff
@@ -129,10 +130,17 @@ public boolean matches(String string) {
...
-            return false;
-
+        int modifier = e.getModifiersEx();
         char keyChar = e.getKeyChar();
+        if ((modifier&(KeyEvent.CTRL_DOWN_MASK|KeyEvent.META_DOWN_MASK)) != 0) {
+            return false;
+        } else if ((modifier & (KeyEvent.ALT_DOWN_MASK)) != 0) {
+            // special case for certain keyboard layouts where diacritics
+            // typed using letter+ALT/OPTION (like Polish Programmer keyboard, the most popular variant in PL).
+            if (!Character.isLetter(keyChar)) {
```
ahadas, line 140: I see two possible issues with this:
1. some of the default shortcuts are set to alt+letter, e.g., alt+D and alt+C
2. users may have configured alt+letter shortcuts
while we can change default shortcuts to address (1), I don't see an easy solution to avoid breaking (2)
maybe we should check here if that combination is set as a shortcut for an action first? while I see that it can be important for users that have Polish Programmer keyboards, I believe those are relatively low number of users that can make sure no shortcuts of this form exist

pskowronek: lemme think about this :)

pskowronek: ad 1 & 2 - pls take look here: https://github.com/mucommander/mucommander/blob/54822d05b53d49301800cfa52a7f857928344c5f/mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTable.java#L1708 - defined shortcuts have already preference

However, here https://github.com/mucommander/mucommander/blob/54822d05b53d49301800cfa52a7f857928344c5f/mucommander-core/src/main/java/com/mucommander/ui/quicklist/item/QuickListDataList.java#L287 there is another use of #isValidQuickSearchInput - and here defined shortcuts are not check, haven't check what QuickListDataList is being used for and whether it matters for this PR.

pskowronek: @ahadas convinced? :)

ahadas: quick lists are a bit different because we don't expect operations to be enabled when a quick list is shown but the first part seems to cover the problematic scenario I had in mind

ahadas (approved):
@pskowronek I've made a slight change in the documentation and moved it to the 'improvements' section since I see this more like an improvement than a bug fix, but other than that it looks great, thanks!

PR #774, opened 2022-10-15
restore timestamp of parent directory after file deleting

### LINKED ISSUE #720: Keep directory timestamp when moving between different mountpoints
Hello,
i have an issue when moving a directory to another directory which is in another mountpoint. The timestamp is the current one and not the original one. When i move files, the timestamp keeps the same. This is what "i" expect for directories too.

Here's a short bash script to create directories and 1 file:
```bash
#!/bin/bash

_dir="2020-08-15_080000"
test -d "$_dir"  &&  rm -r "$_dir"
mkdir -p "$_dir/test"
touch --date="2020-08-15 08:00:00" "$_dir/test/test3"
touch --date="2020-08-15 08:00:00" "$_dir/test"
touch --date="2020-08-15 08:00:00" "$_dir"
```

I tested it at the latest commit hash 6b85a0f1e. I opened muCommander, created the directories by the script above in `/tmp` and then moved it to `/data` which is on a different mountpoint (partition).

Is there a way to keep directory timestamp while using the "Move" option ? At the "Copy" option, the timestamps are kept.

mfilser:
- fixes #720

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/job/impl/MoveJob.java
+++ b/mucommander-core/src/main/java/com/mucommander/job/impl/MoveJob.java
@@ -142,12 +142,11 @@ public class MoveJob extends AbstractCopyJob {
         // - if the 'rename' operation is not supported
         // Note: we want to avoid calling AbstractFile#renameTo when we know it will fail, as it performs some costly
         // I/O bound checks and ends up throwing an exception which also comes at a cost.
-        if(!append && file.getURL().schemeEquals(destFile.getURL()) && file.isFileOperationSupported(FileOperation.RENAME)) {
+        if (!append && file.getURL().schemeEquals(destFile.getURL()) && file.isFileOperationSupported(FileOperation.RENAME)) {
             try {
                 file.renameTo(destFile);
                 return true;
-            }
-            catch(IOException e) {
+            } catch(IOException e) {
                 // Fail silently: renameTo might fail under normal conditions, for instance for local files which are
                 // not located on the same volume.
                 LOGGER.debug("Failed to rename "+file+" into "+destFile+" (not necessarily an error)", e);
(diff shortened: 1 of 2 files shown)


ahadas:
@mfilser thanks for posting this!
in general, I believe we should operate as close as possible to what system tools would do, in this case `mv`. so when I read the issue you described in #720, it made sense to me that we should do as you say, i.e., preserve the original modification date of target directories when they move to a different file system (partition). but when I looked at this PR I could figure out why you've changed the modification date of the *source folders* and now I see - we use cached files so when you do that, the date of the source folder is retrieved before it changes later on when removing its children. so let me post an alternative way to achieve this and then see if that also resolves the issue

ahadas (commented):
@mfilser thanks for posting this!
in general, I believe we should operate as close as possible to what system tools would do, in this case `mv`. so when I read the issue you described in https://github.com/mucommander/mucommander/issues/720, it made sense to me that we should do as you say, i.e., preserve the original modification date of target directories when they move to a different file system (partition). but when I looked at this PR I could figure out why you've changed the modification date of the source folders and now I see - we use cached files so when you do that, the date of the source folder is retrieved before it changes later on when removing its children. so let me post an alternative way to achieve this and then see if that also resolves the issue

ahadas:
@mfilser can you please take a look at [this commit](https://github.com/mucommander/mucommander/pull/774/commits/7f23021657283edd90c851989f5868d5e689e63b) and let me know what you think?

mfilser:
@ahadas 
Thank you for the feedback.
Your commit https://github.com/mucommander/mucommander/pull/774/commits/7f23021657283edd90c851989f5868d5e689e63b works very well too, i tested it right now and the issue is still resolved.

ahadas:
thanks for the confirmation!

Issue #782, opened 2022-10-23
Mark/unmark (+/-) don't work on macOS unless using full KB with numpad or alternative shortcut


pskowronek:
Mark/unmark (+/-) actions don't work on macOS with small keyboard without num-keys. It works only when using +/- from numpad of full layout keyboard. To make it work on small keyboard I had to add alternative shortcut "+" and "-".

I've got two questions:
1) what is the character in the screenshot below?
2) where default shortcuts for these actions are being defined in the project (code)? I don't see them on Action handlers.

![obraz](https://user-images.githubusercontent.com/360665/197396808-f386d4cb-e1f6-4ca5-ae4c-5a138eda467a.png)

ahadas:
> 1. what is the character in the screenshot below?

it's not a character, it's the representation of "NumPad" on macOS - so this thingie with '+' means 'NumPad +'

> 2. where default shortcuts for these actions are being defined in the project (code)? I don't see them on Action handlers.

[here](https://github.com/mucommander/mucommander/blob/master/mucommander-os-api/src/main/java/com/mucommander/desktop/ActionShortcuts.java) is the default ones, and [here](https://github.com/mucommander/mucommander/blob/master/mucommander-os-macos/src/main/java/com/mucommander/desktop/macos/ActionShortcuts.java) we override things for macOS

pskowronek:
> > 1. what is the character in the screenshot below?
> 
> it's not a character, it's the representation of "NumPad" on macOS - so this thingie with '+' means 'NumPad +'

Interesting..... not very intuitive I must say :)

> 
> > 2. where default shortcuts for these actions are being defined in the project (code)? I don't see them on Action handlers.
> 
> [here](https://github.com/mucommander/mucommander/blob/master/mucommander-os-api/src/main/java/com/mucommander/desktop/ActionShortcuts.java) is the default ones, and [here](https://github.com/mucommander/mucommander/blob/master/mucommander-os-macos/src/main/java/com/mucommander/desktop/macos/ActionShortcuts.java) we override things for macOS

Any objections to add "normal" + and - to alternative shortcuts (as default ones)?

PR #785, opened 2022-10-26
Added alternative default shortcuts for Mark/Unmark for keyboard w/o …

pskowronek:
Added alternative default shortcuts for Mark/Unmark for keyboard w/o numpad (#782)

### DIFF
--- a/mucommander-os-api/src/main/java/com/mucommander/desktop/ActionShortcuts.java
+++ b/mucommander-os-api/src/main/java/com/mucommander/desktop/ActionShortcuts.java
@@ -284,10 +284,12 @@ public class ActionShortcuts {
             return KeyStroke.getKeyStroke(KeyEvent.VK_LEFT, 0);
         case GoToRoot:
             return KeyStroke.getKeyStroke(KeyEvent.VK_BACK_SLASH, KeyEvent.CTRL_DOWN_MASK);
+        case InvertSelection:
+            return KeyStroke.getKeyStroke(KeyEvent.VK_8, KeyEvent.CTRL_DOWN_MASK | KeyEvent.SHIFT_DOWN_MASK);
         case MarkExtension:
-            return KeyStroke.getKeyStroke(KeyEvent.VK_EQUALS, 0);
+            return KeyStroke.getKeyStroke(KeyEvent.VK_EQUALS, KeyEvent.CTRL_DOWN_MASK);
         case MarkGroup:
-            return KeyStroke.getKeyStroke(KeyEvent.VK_EQUALS, KeyEvent.SHIFT_DOWN_MASK);
+            return KeyStroke.getKeyStroke(KeyEvent.VK_EQUALS, KeyEvent.CTRL_DOWN_MASK | KeyEvent.SHIFT_DOWN_MASK);
         case MarkSelectedFile:
             return KeyStroke.getKeyStroke(KeyEvent.VK_INSERT, 0);
         case NextTab:
@@ -305,7 +307,7 @@ public class ActionShortcuts {
         case SwitchActiveTable:
             return KeyStroke.getKeyStroke(KeyEvent.VK_TAB, KeyEvent.SHIFT_DOWN_MASK);
         case UnmarkGroup:
-            return KeyStroke.getKeyStroke(KeyEvent.VK_MINUS, 0);
+            return KeyStroke.getKeyStroke(KeyEvent.VK_MINUS, KeyEvent.CTRL_DOWN_MASK);
         default:
             return null;
         }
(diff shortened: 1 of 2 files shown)


mucommander-os-api/src/main/java/com/mucommander/desktop/ActionShortcuts.java
```diff
@@ -284,6 +284,10 @@ public KeyStroke getDefaultAltKeyStroke(ActionType actionId) {
             return KeyStroke.getKeyStroke(KeyEvent.VK_LEFT, 0);
         case GoToRoot:
             return KeyStroke.getKeyStroke(KeyEvent.VK_BACK_SLASH, KeyEvent.CTRL_DOWN_MASK);
+        case MarkExtension:
+            return KeyStroke.getKeyStroke(KeyEvent.VK_EQUALS, 0);
```
ahadas, line 288: wouldn't this mean that we won't be able to trigger quick-search by pressing '='? same for '-' that is handled below

pskowronek: by quick-search you mean moving the cursor when you start typing its name?
If so, then yes, those '=' and '-' (and '+') won't work for quick search initial character. Additional remark, making non-numpad '+' and '-' work as their numpad counter-parts will make it more consistent (especially than in muC main menu under `Mark` you can see '+' and '-' for Mark/Unmark actions and you wonder why they don't work if you don't have numpad /I guess pretty common nowadays on laptops/)

ahadas: > by quick-search you mean moving the cursor when you start typing its name? If so, then yes, those '=' and '-' (and '+') won't work for quick search initial character.

yep, that's what I meant - so it would be better to choose alternative shortcuts that don't limit this functionality

> Additional remark, making non-numpad '+' and '-' work as their numpad counter-parts will make it more consistent (especially than in muC main menu under `Mark` you can see '+' and '-' for Mark/Unmark actions

yeah, that depends on the operating system - we saw that weird representation of numpad in macOS which may confuse users but in fedora I believe it's clearer as, e.g., it's rendered as "NumPad +" for "Mark files..."

> and you wonder why they don't work if you don't have numpad\ /I guess pretty common nowadays on laptops/)

right, and not only on laptops - for instance, I use the Apple Magic Keyboard that doesn't have numpad. I agree we should avoid having actions that are only correlated with shortcuts that include numpad keys

pskowronek: > yep, that's what I meant - so it would be better to choose alternative shortcuts that don't limit this functionality

I haven't seen in my life a file or folder with leading = or +, and having a different shortcut without +/=/- won't be intuitive. If somebody wanted to quick search for such exotic filenames then they can remove such shortcuts from defaults.

> > Additional remark, making non-numpad '+' and '-' work as their numpad counter-parts will make it more consistent (especially than in muC main menu under `Mark` you can see '+' and '-' for Mark/Unmark actions
> 
> yeah, that depends on the operating system - we saw that weird representation of numpad in macOS which may confuse users but on fedora I believe it's clearer as, e.g., the shortcut of "Mark files..." is rendered as "NumPad +"

This doesn't solve problem for ppl not having numpads :)

> 
> > and you wonder why they don't work if you don't have numpad\ /I guess pretty common nowadays on laptops/)
> 
> right, and not only on laptops - for instance, I use the Apple Magic Keyboard that doesn't have numeric keypad. I agree we should avoid having actions that are only correlated with shortcuts that include numpad keys

Let's figure out something then :)

ahadas: @pskowronek the way quick-search works doesn't force to match the filename right from the start, I can type "-tmp" and all folders that match ".*-tmpdir.*" would match, so the best practice would be not to use keys that can be part of filenames as shortcuts - how about adding 'ALT' to all those shortcuts?

pskowronek: Tried with ALT:

1) ALT+Shift+=  produces this letter on mac: https://en.wiktionary.org/wiki/%C4%AB - as it is a letter, then quick search is engaged 
2) ALT+= - it looks like single file selection is engaged

To avoid the mess, I've tried with CTRL - no issues on mac. I will provide update in a minute (plus Ctrl+Shift+8 -> * i.e. invert selection). Pls tell me what you think.

pskowronek:
PR updated - Using Ctrl with +/-/* for non-numpad keyboards

mucommander-os-api/src/main/java/com/mucommander/desktop/ActionShortcuts.java
ahadas: looks good, did you check that those shortcuts don't conflict with existing ones?

pskowronek: checked if sth reacts to those shortcuts and also whether if they conflict if existing ones by manually adding them in Preferences - nothing came up.

ahadas (approved):
thanks @pskowronek!

ahadas:
Version 1.2.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.2.0-1)

PR #795, opened 2022-11-26
Terminal fixes: hide initial cd directory command, hide terminal when…

pskowronek:
Terminal fixes: 
- hide initial cd directory command
- hide terminal when it is closed via `exit`/Ctrl-D command line and another key is pressed


ahadas (approved):
nice, thanks!

Issue #801, opened 2022-12-09
Components' update


aashipov:
mucommander relies on a bunch of opensource free-licensed libraries. A lot of 'collective effort' is invested in those components development. For mucommander to benefit from that it seems wise to update project dependencies

PR #802, opened 2022-12-09
#801 Components' update

aashipov:


ahadas (approved):
the upgrade to OSGi 8 looked a bit risky but it seems to work well

ahadas:
@aashipov seems that in important functionality for development, listing the bundles using `ps`, doesn't work with v2.0 of osgi-run so I reverted that part (that's the challenge with such PRs, to see the hidden impact of these changes)

Issue #811, opened 2022-12-23
Opening links from viewer/editor does not work reliably


Andreas0602:
On macOS Ventura, opening links from the new viewer/editor does not work reliably. In rare cases, clicking on a link like www.google.com while pressing the command key will open the link in Safari, but in ~ 95% of all attempts nothing will happen.

ahadas:
it doesn't reproduce for me on macOS Monterey (on the contrary, the link seems to open too quick - on mouseover after clicking the command key, before clicking the link). @pskowronek can you please look into this?

Andreas0602:
Yes, sometimes the link even opens after pressing and releasing the command key, without a mouse click. But in most of the cases, it doesn't do anything at all.

pskowronek:
> it doesn't reproduce for me on macOS Monterey (on the contrary, the link seems to open too quick - on mouseover after clicking the command key, before clicking the link). @pskowronek can you please look into this?

as it comes "too quick" - I have this, too, but I thought it is by design in RSyntaxTextArea - I will check in their vanilla example how it behaves.

>  but in ~ 95% of all attempts nothing will happen

I don't have Ventura anywhere :/ Can somebody verify on Ventura if it is the problem?

pskowronek:
btw, @Andreas0602 - for opening links I've used muC existing code - can you check if in AboutDialog clicking `Go to website` opens browser reliably?

Andreas0602:
Clicking `Go to website` in the About dialog does open the browser reliably.

pskowronek:
I've just pushed PR
- @ahadas - pls check key sequence
- @Andreas0602 - pls check reliability of opening the links

Andreas0602:
@pskowronek, I tested your change. I still need several attempts to open a link, but apparently fewer than before. After opening the link, the mouse cursor will change from an arrow to a pointing hand, and it's not possible to change it back except by closing the window. So, the behavior has slightly improved, but it's still not okay.

Andreas0602:
Okay, opening a link does work reliably now if I do it the following way:

step 1: place the mouse cursor over the link
step 2: press the command key and keep it pressed
step 3: slightly move the mouse cursor -- as a result, the mouse cursor will change from an arrow to a pointing hand
step 4: left mouse click

It doesn't work without step 3.
But it should work without step 3.

pskowronek:
@Andreas0602 - I cannot reproduce this on my mac (Catalina) :/

[...]

PR #817, opened 2022-12-24
Handle link scanning mask pressing/releasing

ahadas:


Andreas0602:
Works well for me. From my point of view, this is the final solution.

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -98,7 +103,28 @@ public Insets getInsets() {
...
+                if (textArea.getHyperlinksEnabled() && (e.getModifiersEx() & linkScanningMask) == linkScanningMask)
+                    fakeMouseMove(e);
+            }
+
+            @Override
+            public void keyReleased(KeyEvent e) {
+                fakeMouseMove(e);
+            }
 
+            private void fakeMouseMove(KeyEvent e) {
```
pskowronek, line 120: If I understand it correctly, this is an attempt to workaround the problem on Ventura?
If so, wouldn't be helpful to add a comment with a link to #811 ?

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -98,7 +103,28 @@ public Insets getInsets() {
                 return new Insets(0, 3, 4, 3);
             }
         };
+        int linkScanningMask = OsFamily.MAC_OS.isCurrent() ? InputEvent.META_DOWN_MASK : InputEvent.CTRL_DOWN_MASK;
```
pskowronek, line 106: What about moving the whole code inside this IF:
```
if (DesktopManager.canBrowse()) {
            textArea.addHyperlinkListener((event) -> {
```

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -98,7 +103,28 @@ public Insets getInsets() {
...
+        textArea.setLinkScanningMask(linkScanningMask);
+        textArea.addKeyListener(new KeyAdapter() {
+            @Override
+            public void keyPressed(KeyEvent e) {
+                if (textArea.getHyperlinksEnabled() && (e.getModifiersEx() & linkScanningMask) == linkScanningMask)
```
pskowronek, line 111: and this clause `textArea.getHyperlinksEnabled()` would not be necessary

pskowronek (commented):
added some review comments

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
ahadas: it happened to me on Monterey - one had to move the cursor while meta/ctrl was pressed in order to get the 'hand' cursor and to have the url rendered as a link, which make sense when you look at the code inside `RSyntaxTextArea` that is responsible for that. so that's not a workaround but rather an enhancement so one wouldn't have to move the mouse - the aforementioned behavior would happen also when the cursor is placed on a link and then meta/ctrl is pressed

pskowronek (approved):
looks good

Andreas0602:
Hmm, with the commit [a6861c5], I still see the following problem:

>Okay, opening a link does work reliably now if I do it the following way:
>
>step 1: place the mouse cursor over the link
>step 2: press the command key and keep it pressed
>step 3: slightly move the mouse cursor -- as a result, the mouse cursor will change from an arrow to a pointing hand
>step 4: left mouse click
>
>It doesn't work without step 3.
>But it should work without step 3.

Is this supposed to be fixed with commit [a6861c5]? It was fixed when I commented `Works well for me. From my point of view, this is the final solution`, but now it's not fixed (for me).

ahadas:
@Andreas0602 yes but it's better to test it also with #818  - could you please try with the nightly build that now includes both?

Andreas0602:
@ahadas, With the nightly build, the handling of links is fine. Also the default cursor type (I-beam instead of arrow) is good. The following behavior would be even better: If the cursor is located over text then it's an I-beam, and if the cursor is not located over text (ie. over empty space) then it's an arrow. This GitHub page in the browser, for example, behaves like that.

ahadas:
thanks for the confirmation @Andreas0602 !

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
pskowronek: @ahadas - have you tried to run RSyntaxTextArea demo app and see if you reproduce this? If so, then maybe its worth to report an bug to RSyntaxTextArea project?

ahadas: nope, I didn't try with the demo app but I'd be really surprised if it works there - filing a bug or even contributing the change to the RSyntaxTextArea project makes a lot of sense

PR #818, opened 2022-12-24
Fix initial focus of text editor/viewer

ahadas:


pskowronek:
hmm, it works :) but I don't know why! What magic is calling this new `requestFocus`?

pskowronek (approved):
looks good

ahadas:
> hmm, it works :) but I don't know why! What magic is calling this new `requestFocus`?

:)
yeah, took me some time to find it - see [this](https://github.com/mucommander/mucommander/blob/1.1.0/mucommander-core/src/main/java/com/mucommander/ui/viewer/FileFrame.java#L100)

PR #820, opened 2022-12-24
Viewer/Editor TEXT_CURSOR as default cursor (still there's problem on…

pskowronek:
TEXT_CURSOR as default cursor (still there's problem on initial open - see TODO) #811

### DIFF
--- a/mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
+++ b/mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
@@ -18,6 +18,7 @@
 package com.mucommander.viewer.text;
 
 import java.awt.Color;
+import java.awt.Cursor;
 import java.awt.Font;
 import java.awt.Insets;
 import java.awt.Point;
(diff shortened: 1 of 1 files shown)


mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -190,6 +191,9 @@ public void mouseWheelMoved(MouseWheelEvent e) {
                 }
             }
         });
+        // Setting TEXT_CURSOR for text area regardless if it is editable or readonly - for both it makes sense.
+        // TODO still not ideal - initially most often it opens with DEFAULT_CURSOR, moving to line numbers panel and back to editor it shows TEXT_CURSOR properly
```
ahadas, line 195: it works well on Linux (Fedora), I'll check on my macbook

ahadas: @pskowronek just a quick update on this - it reproduces on my macbook but couldn't pinpoint the problem yet

ahadas: ok, this also happens with a simple JFrame and JTextArea on macOS so it doesn't look like a result of our code - maybe it would be fixed with a newer version of Java
but I don't see a reason to block this PR on that

ahadas (approved):

PR #821, opened 2022-12-25
Viewer/Editor make utf-8 default instead of ISO-8859-1

pskowronek:
Viewer/Editor make utf-8 default instead of ISO-8859-1.

Use case:
- open in Editor any java file in muC
- in Editor menu -> encoding you can see `ISO-8859-1`
- start typing some UTF-8 national characters or paste UTF-8 emoticon
- save
- re-open - `ISO-8859-1` is still chosen and recently modified text is broken

What I propose here is to use UTF-8 a default when `ISO-8859-1` detected. Other quick idea is to use system default charset instead of hard-coded UTF-8.


ahadas (approved):
I think it's a better approach that using the system default charset - UTF-8 is an extension of ISO-8859-1 so I don't see any issue in using it instead

Issue #565, opened 2022-01-22
Total progress bar does not work


Andreas0602:
Under macOS Monterey, when copying large files, the total progress bar always shows "0% - ?" instead of the real progress. To reproduce the issue, follow these steps:

1. Create two separate directories: A and B
2. Inside of directory A, create another directory C
3. Copy a large file (which is ~ 20 GB or larger) into directory C
4. In the left pane of muCommander, go to directory A
5. In the right pane of muCommander, go to directory B
6. Make the left pane of muCommander the active one and select directory C
7. Press F5 and then press the "Copy" button

As a result, muCommander will copy C from A to B. While the copying is ongoing, the total progress bar will show "0% - ?" all the time.

Andreas0602:
Other file managers for macOS, for example Commander One or zCommander, do not have this issue.

Andreas0602:
In fact, I'm not aware of any file manager on any platform that doesn't show the total progress when copying.

PR #823, opened 2022-12-26
Dock/TaskBar icon status with jobs progress updates

pskowronek:
Added FileJobs' status updates on App Icon - under macOS in Dock or in Taskbar in Windows/Ubuntu (Unity).
- progress (probably not calculated accurately - probably I should use `getTotalPercentDone` instead of `getJobProgress.getFilePercentInt`)
- badge number - number of jobs (all i.e. foreground and background jobs)
- icon bouncing when user attention is required (file exists etc)

Screenshot:
muc-icon

This is PoC - only checked on macOS Catalina.

All comments are welcomed.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
@@ -79,7 +79,7 @@ public class MainFrame extends JFrame implements LocationListener {
 
     private FolderPanel leftFolderPanel;
     private FolderPanel rightFolderPanel;
-	
+
     private FileTable leftTable;
     private FileTable rightTable;
     
@@ -97,10 +97,10 @@ public class MainFrame extends JFrame implements LocationListener {
 
     /** Status bar instance */
     private StatusBar statusBar;
-	
+
     /** Command bar instance */
     private CommandBar commandBar;
-	
+
     /** Is no events mode enabled ? */
     private boolean noEventsMode;
 
@@ -227,22 +227,24 @@ public class MainFrame extends JFrame implements LocationListener {
         // Note: the vertical/horizontal terminology used in muCommander is just the opposite of the one used
         // in JSplitPane which is anti-natural / confusing.
         splitPane = new ProportionalSplitPane(this,
-                MuSnapshot.getSnapshot().getVariable(MuSnapshot.getSplitOrientation(0), MuSnapshot.DEFAULT_SPLIT_ORIENTATION).equals(MuSnapshot.VERTICAL_SPLIT_ORIENTATION) ?
-                                              	JSplitPane.HORIZONTAL_SPLIT:JSplitPane.VERTICAL_SPLIT,
-                                              false,
-                                              MainFrame.this.leftFolderPanel,
-                                              MainFrame.this.rightFolderPanel) {
-        	// We don't want any extra space around split pane
-        	@Override
-        	public Insets getInsets() {
-        		return new Insets(0, 0, 0, 0);
-        	}
+                MuSnapshot.getSnapshot().getVariable(
+                        MuSnapshot.getSplitOrientation(0),
+                        MuSnapshot.DEFAULT_SPLIT_ORIENTATION).equals(MuSnapshot.VERTICAL_SPLIT_ORIENTATION) ?
+                        JSplitPane.HORIZONTAL_SPLIT:JSplitPane.VERTICAL_SPLIT,
+                        false,
+                        MainFrame.this.leftFolderPanel,
+                        MainFrame.this.rightFolderPanel) {
+            // We don't want any extra space around split pane
+            @Override
+            public Insets getInsets() {
+                return new Insets(0, 0, 0, 0);
+            }
         };
 
         // Remove any default border the split pane has
         splitPane.setBorder(null);
 
-        // Adds buttons that allow to collapse and expand the split pane in both directions
+        // Add buttons that allow to collapse and expand the split pane in both directions
         splitPane.setOneTouchExpandable(true);
 
         // Disable all the JSplitPane accessibility shortcuts that are registered by default, as some of them
@@ -259,7 +261,7 @@ public class MainFrame extends JFrame implements LocationListener {
         // Add status bar
         this.statusBar = new StatusBar(this);
         southPanel.add(statusBar);
-		
+
         // Show command bar only if it hasn't been disabled in the preferences
         this.commandBar = new CommandBar(this);
         // Note: CommandBar.setVisible() has to be called no matter if CommandBar is visible or not, in order for it to be properly initialized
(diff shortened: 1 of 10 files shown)


mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -114,24 +119,58 @@ public class MainFrame extends JFrame implements LocationListener {
...
+            icons.add(IconManager.getIcon(IconManager.MUCOMMANDER_ICON_SET, "icon32_24.png").getImage());
+            icons.add(IconManager.getIcon(IconManager.MUCOMMANDER_ICON_SET, "icon48_24.png").getImage());
+            icons.add(IconManager.getIcon(IconManager.MUCOMMANDER_ICON_SET, "icon128_24.png").getImage());
+            icons.add(IconManager.getIcon(IconManager.MUCOMMANDER_ICON_SET, "icon256_24.png").getImage());
+
+            setIconImages(icons);
+        }
 
-        java.util.List<Image> icons = new Vector<Image>();
+        JobsManager.getInstance().addJobListener(new JobListener() {
```
pskowronek, line 140: Is it a good place to add such a listener?

mucommander-core/src/main/java/com/mucommander/ui/notifier/NotifierProvider.java
```diff
@@ -98,6 +98,7 @@ public static void displayBackgroundNotification(final NotificationType notifica
                 return;
             }
 
+            DesktopManager.requestUserAttention();  // bounce app icon if supported by OS
```
pskowronek, line 101: Is it a good place?

ahadas:
interesting :) I need to take a better look at the details though

pskowronek:
rebased

pskowronek:
> interesting :) I need to take a better look at the details though

Can you take a look? Your input will be valuable, then I would try to prepare a proper PR.

ahadas:
> @ahadas Can you take a look? Your input will be valuable, then I would try to prepare a proper PR.

sure, sorry for not getting back to it sooner - I'll try to do it this week

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
ahadas: well, generally I don't see the benefit of using multiple frames now that we have tabs and the ability to execute operations in the background but as this option of having multiple frames still exists, I'd say we shouldn't add a listener per frame. wouldn't it make more sense to integrate it into the `com.mucommander.ui.notifier` package?

mucommander-core/src/main/java/com/mucommander/ui/notifier/NotifierProvider.java
ahadas: yeah, this seems like a good place for this. I think this package might also be a good place for the other parts - it seems to be that all the changes here are related to notifications

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
```diff
@@ -155,4 +156,46 @@ public boolean isApplication(AbstractFile file) {
     public ActionShortcuts getActionShortcuts() {
         return new ActionShortcuts();
     }
+
+    @Override
+    public boolean setIconBadgeNumber(int number) {
+        boolean opSuccessful = false;
+        if (Taskbar.isTaskbarSupported() &&
+                Taskbar.getTaskbar().isSupported(Taskbar.Feature.ICON_BADGE_TEXT)) {
+            if (number >= 0) {
```
ahadas, line 165: shouldn't it be `number > 0`?

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
```diff
@@ -155,4 +156,46 @@ public boolean isApplication(AbstractFile file) {
...
+                Taskbar.getTaskbar().setProgressValue(-1);
+            }
+            opSuccessful = true;
+        }
+        return opSuccessful;
+    }
+
+    @Override
+    public boolean requestUserAttention() {
+        boolean opSuccessful = false;
```
ahadas, line 192: how about inline this variable? the return value is already documented (same for those variables in the methods above)

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
```diff
@@ -155,4 +156,46 @@ public boolean isApplication(AbstractFile file) {
...
+        if (Taskbar.isTaskbarSupported() &&
+                Taskbar.getTaskbar().isSupported(Taskbar.Feature.USER_ATTENTION)) {
+            // TODO check if we need to "unrequest" attention
```
ahadas, line 195: yeah and also we may want to set the window, in case there are multiple main frames (but again, I'm not sure it's that important - I don't find much use for using more than one frame)

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -114,24 +119,59 @@ public class MainFrame extends JFrame implements LocationListener {
...
+                    // Update icon every 1s
+                    if (lastUpdate + 1000L < System.currentTimeMillis()) {
+                        lastUpdate = System.currentTimeMillis();
+                        long sum = 0;
+                        int jobsCount = 0;
+                        for (FileJob job : jobs) {
+                            sum += job.getJobProgress().getTotalPercentInt();
+                            jobsCount++;
+                        }
+                        DesktopManager.setIconBadgeNumber(jobs.size());
```
ahadas, line 166:
```suggestion
                        DesktopManager.setIconBadgeNumber(jobsCount);
```

mucommander-os-macos/src/main/java/com/mucommander/desktop/macos/OSXDesktopAdapter.java
```diff
@@ -242,4 +243,40 @@ private static String getMacOsUserShell() {
         }
         return result;
     }
+
+    @Override
+    public boolean setIconBadgeNumber(int number) {
+        // Hmm, Taskbar.isTaskbarSupported returns false on my macOS 10.15.7 and java 14.0.2
```
ahadas, line 249: as long as the dmg we publish are supported on those older versions that's ok - it would be interesting to see if 1.3.0 with Java 20 would work on macOS 10.15.7

ahadas (commented):
@pskowronek it may seems like a high number of comments but most of them are minor, this change looks nice overall - but as you commented inside, let's find the right place to have this bridge from the jobs to the icon's badge and progress

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
pskowronek: Matter of taste, I like having a variable that is being returned. If you insist I can change, I prefer the way it is.

pskowronek:
PR updated to address all minor comments (that I've marked as resolved).

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
ahadas: yeah, that's subjective but while I was trying to think of a better name to it, it occurred to me that removing it would be simpler.. the name `opSuccessful` suggests that the operation can complete successfully or fail while the value of this field rather indicates whether the operation was triggered - so I'd say it should be renamed to something like 'opSupportedByAwt' or 'awtActionTriggered'

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
pskowronek: > yeah and also we may want to set the window, in case there are multiple main frames

if I get this right, then there's no way to specify main frame while invoking Taskbar.getTaskbar().requestUserAttention. On macOS there's one dock icon for multiple main frames, don't remember how it is on Windows or Linux.

ahadas: right, but I saw requestWindowUserAttention - not sure if those two methods can go well together

pskowronek: ah yes, however, job listeners don't have a clue what main frame executed them so, I would try to go with a simple `requestUserAttention` and see if it is OK for users?

pskowronek:
PR updated - jobs listener registration moved to NotifierProvider

mucommander-os-macos/src/main/java/com/mucommander/desktop/macos/OSXDesktopAdapter.java
pskowronek: so, with Java 20 `Taskbar.isTaskbarSupported` returns true and it works with both badges and progressbar (but with workaround - just pushed a commit to address that, apparently a bug in AWT that exists up to 21).

pskowronek:
> On macOS Ventura, I only see a job counter, but I don't see a progress bar. By the way, this would be a great opportunity to address #565. 😀
> 
> Dock

I've just pushed a commit with workaround for that. @Andreas0602 can you check it?

pskowronek:
> > On macOS Ventura, I only see a job counter, but I don't see a progress bar. By the way, this would be a great opportunity to address #565. 😀
> > Dock
> 
> I've just pushed a commit with workaround for that. @Andreas0602 can you check it?

@ahadas / @Andreas0602  or anyone - do you have account on bugs.openjdk.org to add a comment to https://bugs.openjdk.org/browse/JDK-8300037 ?

pskowronek:
> By the way, here's a screenshot of the progress bar of NewsLazer at about 75 %. I must say that it looks better.

ours or theirs? :)

pskowronek:
Matter of taste :) However, ours is a native one, and theirs is custom painted and updated icon.

jre
ahadas, line 1: need to drop this change

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
ahadas: > > @ahadas / @Andreas0602 or anyone - do you have account on bugs.openjdk.org ...
> 
> No, I haven't.

Me neither

package/readme.txt
```diff
@@ -36,6 +36,7 @@ New features:
 - Undo and redo actions for the internal binary editor.
 - Additional search options for the internal text viewer/editor: case-sensitivity, whole-words, regular expressions, forward/backward.
 - Browsing and extracting multi-volume RAR 5+ files.
+- Dock/TaskBar icon showing jobs' progress updates
```
ahadas, line 39: how about:
```suggestion
- Dock/TaskBar icon shows the amount of active jobs and their progress
```

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
```diff
@@ -155,4 +157,58 @@ public boolean isApplication(AbstractFile file) {
...
+
+            // Apparently there's a problem with AWT with above code on macOS (https://bugs.openjdk.org/browse/JDK-8300037), so:
+            // work around starts:
+            EventQueue.invokeLater(() -> {
+                Taskbar.getTaskbar().setIconImage(Taskbar.getTaskbar().getIconImage());
+                if (progress >= 0 && progress <= 100) {
+                    Taskbar.getTaskbar().setProgressValue(progress);
+                } else {
+                    Taskbar.getTaskbar().setProgressValue(-1);
+                }
```
ahadas, line 195: we need to keep the duplicate code above then?

ahadas (commented):
minor comments, other than that lgtm

ahadas:
> > @ahadas / @Andreas0602 or anyone - do you have account on bugs.openjdk.org ...
> 
> No, I haven't.

me neither

package/readme.txt
pskowronek: good one

pskowronek:
PR updated

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
```diff
@@ -155,4 +157,54 @@ public boolean isApplication(AbstractFile file) {
...
+    @Override
+    public boolean setIconProgress(int progress) {
+        boolean opSupportedByAwt = false;
+        if (Taskbar.isTaskbarSupported() &&
+                Taskbar.getTaskbar().isSupported(Taskbar.Feature.PROGRESS_VALUE)) {
+            if (progress >= 0 && progress <= 100) {
+                Taskbar.getTaskbar().setProgressValue(progress);
+            } else {
+                Taskbar.getTaskbar().setProgressValue(-1);
+            }
```
ahadas, line 185: @pskowronek what I meant before is whether we can move the 5 lines above to the place that is invoked later below

jre
pskowronek: I should remove it from PR, right?

mucommander-os-api/src/main/java/com/mucommander/desktop/DefaultDesktopAdapter.java
pskowronek: ok, PR updated

jre
ahadas: I've pushed a commit that reverts that change. it seems like the 'jre' submodule is not up-to-date, did you try `git submodule update`?

ahadas (approved):
thanks @pskowronek !
note that [I commented on the 'jre' thread](https://github.com/mucommander/mucommander/pull/823#discussion_r1189121419)  - I think you just need to update the submodule with `git submodule init`

jre
pskowronek: hm, it does nothing, the same with `git submodule init`. Previously I also tried `git submodule foreach git reset --hard` - it did something, but..... heck, still I don't know how come it went to commit and how it should be reverted.

ahadas: ah yeah, actually it means the submodule was updated because otherwise it would't have detected a change in 'jre' - did you rebase the root repo? it seems like you are missing [this commit](https://github.com/mucommander/mucommander/commit/14f07794ecd9c704cfd51754f3586ec94ccab799)

ahadas:
Version 1.3.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.3.0-1)

Issue #835, opened 2023-01-04
Text editor should create utf-8 files without BOM by default


Andreas0602:
When you create a shell script and try to run it, you'll get an error like "command not found". To reproduce the issue, follow these steps:

1. Press Shift+F7 and create a new file named "test".
2. Select the file "test" and press F4. This will open the file in the internal editor.
3. In the internal editor, enter the text "echo Hello world"
4. Press ESC. When prompted, press "Save" to save the file.
5. Press F12 to open the terminal.
6. Execute the following command: sh ./test

To fix this issue, the text editor should create utf-8 files without BOM by default.

pskowronek:
I confirm, will take a look.

pskowronek:
This behavior was also present in muC Version: 1.1.0 (when UTF-8 was set as encoding).

pskowronek:
Since BOM for UTF-8 is not required and in the wild is uncommon, plus having [this](https://unicode.org/faq/utf_bom.html#utf8-2) information I decided to override BOMWriter#bomWriteChecked init value so it will avoid writing BOM for UTF-8.

See upcoming PR

pskowronek:
PR provided #836

ahadas:
I have a vague memory of a thread on dropping BOM, which didn't make sense back then, but for specifically for UTF-8 it sounds ok - so this approach is fine by me, I'll comment on implementation details in the PR

PR #836, opened 2023-01-04
Editor Fix to avoid writing BOM for new UTF-8 files

pskowronek:
Editor: Fix to avoid writing BOM for new UTF-8 files - #835

### DIFF
--- a/mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditor.java
+++ b/mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditor.java
@@ -201,7 +201,13 @@ class TextEditor extends BasicFileEditor implements DocumentListener, EncodingLi
     }
 
     private void write(OutputStream out) throws IOException {
-        textEditorImpl.write(new BOMWriter(out, textViewerDelegate.getEncoding()));
+        boolean isUtf8 = "UTF-8".equalsIgnoreCase(textViewerDelegate.getEncoding());
+        textEditorImpl.write(new BOMWriter(out, textViewerDelegate.getEncoding()) {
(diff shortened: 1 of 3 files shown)


mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditor.java
```diff
@@ -201,7 +201,13 @@ void loadDocument(InputStream in, String encoding, DocumentListener documentList
...
 
     private void write(OutputStream out) throws IOException {
-        textEditorImpl.write(new BOMWriter(out, textViewerDelegate.getEncoding()));
+        boolean isUtf8 = "UTF-8".equalsIgnoreCase(textViewerDelegate.getEncoding());
+        textEditorImpl.write(new BOMWriter(out, textViewerDelegate.getEncoding()) {
+            {
+                // Don't write BOM for utf-8 - see: https://github.com/mucommander/mucommander/issues/835
+                bomWriteChecked = isUtf8;
+            }
+        });
```
ahadas, line 210: so why not to just use OutputStreamWriter in case of UTF-8?

ahadas (changes requested):
it is also worth deprecating BOMConstants#SUPPORTED_BOMS#UTF8_BOM with a proper comment

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditor.java
pskowronek: fair remark :)

pskowronek:
PR updated

ahadas (approved):

Issue #838, opened 2023-01-06
Pressing F3 in viewer does not find search text


Andreas0602:
1. Create a text file containing the word `Test`.
2. In muCommander, go to `File` > `Find...`
3. In the field `Containing text`, enter `Test`. Leave the option `Case sensitive` deactivated.
4. Start the search. muCommander will find the text file.
5. Select the text file and press F3. This will open the text file in the viewer.
6. Press F3 again.

**Expected result:** The word `Test` in the file will be highlighted.
**Actual result:** The word `Test` in the file will not be highlighted. Instead, you can hear a beep sound.

When you search for `test` instead of `Test`, the issue will not occur.

pskowronek:
This behavior existed in previous muC versions.

In Viewer/Editor search is being done in lower-case, whereas file search is done either in case sensitive or insensitive mode and search string is being passed to Editor w/o information in what mode it was.
I've provided a fix / workaround so search string is converted to lowercase. The real fix what be to add case sensitive or insensitive modes to Viewer/Editor search (Dialog and search mechanism plus pass the mode from File search).

ahadas:
thanks @pskowronek! yeah, adding case-sensitive search is missing - I didn't plan to add it because the text viewer/editor was pretty basic and the assumption was that users that want something more advanced would anyway configure an external application, but with the recent changes you've contributed - I don't mind adding it

PR #839, opened 2023-01-07
Viewer/Editor fix/workaround F3 for file search #838

pskowronek:
Viewer/Editor fix or rather a workaround F3 for file search #838.


ahadas (commented):
yeah, as you pointed out in #838 , the text search is case insensitive and that wasn't taken into account when using the text from the file search - which needs to be converted to lower case. when we do the conversion of lower case in `doSearch`, we no longer need to do that in `find` - I'll add another commit to drop it from there

ahadas (approved):

PR #873, opened 2023-02-25
Eager loading of snapshot properties

ahadas:
Previously, with lazy loading, snapshot properties could have been dropped in the following flow:
1. Update a snapshot configuration
2. Quit the application so the snapshot configuration would be persisted to snapshot.xml
3. Start the application
4. Close the application
5. Start the application
6. The snapshot configuration was reverted to its default value

The reason why the snapshot configuration was removed is because it wasn't loaded between step 3 and step 4 and therefore when closing the application, we cleared the snapshot configuration that is going to be persisted (in order to get rid of properties that were dropped) and then, when updating the snapshot configuration, we didn't get updated value of the snapshot configuration from step 1.

The solution is to override the `read` method of `MuSnapshotable` to make sure the snapshot configuration are loaded on startup.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/search/SearchSnapshot.java
+++ b/mucommander-core/src/main/java/com/mucommander/search/SearchSnapshot.java
@@ -16,33 +16,27 @@
  */
 package com.mucommander.search;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static com.mucommander.snapshot.MuSnapshot.SEARCH_SECTION;
 
 import com.mucommander.commons.conf.Configuration;
-import com.mucommander.snapshot.MuSnapshot;
 import com.mucommander.snapshot.MuSnapshotable;
 
-public class SearchSnapshot implements MuSnapshotable {
-    private static final Logger LOGGER = LoggerFactory.getLogger(SearchSnapshot.class);
-
-    @Override
-    public void read(Configuration configuration) {
-        LOGGER.info("Loading snapshot configuration for " + SearchSnapshot.class);
-        for (var property : SearchProperty.values()) {
-            var prefKey = MuSnapshot.SEARCH_SECTION + "." + property.getKey();
-            property.setValue(MuSnapshot.getSnapshot().getVariable(prefKey, property.getValue()));
-        }
+/**
+ * Snapshot preferences for file-search
+ * @author Arik Hadas
+ */
+public class SearchSnapshot extends MuSnapshotable<SearchProperty> {
+    public SearchSnapshot() {
+        super(SearchProperty::values,
+                SearchProperty::getValue,
+                SearchProperty::setValue,
+                pref -> pref.getKey() != null ? SEARCH_SECTION + "." + pref.getKey() : null);
     }
 
     @Override
-    public void write(Configuration configuration) {
-        for (var property : SearchProperty.values()) {
-            if (!property.isDefault()) {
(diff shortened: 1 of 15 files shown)


mucommander-core/src/main/java/com/mucommander/search/SearchSnapshot.java
```diff
@@ -16,11 +16,24 @@
...
 import com.mucommander.commons.conf.Configuration;
 import com.mucommander.snapshot.MuSnapshot;
 import com.mucommander.snapshot.MuSnapshotable;
 
 public class SearchSnapshot implements MuSnapshotable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SearchSnapshot.class);
+
+    @Override
+    public void read(Configuration configuration) {
+        LOGGER.info("Loading snapshot configuration for " + SearchSnapshot.class);
```
pskowronek, line 31: LOGGER.info("Loading snapshot configuration for: {}" , SearchSnapshot.class); ?

mucommander-core/src/main/java/com/mucommander/ui/viewer/EditorSnapshot.java
```diff
@@ -16,18 +16,39 @@
...
 
 /**
  * Snapshot preferences for editor.
  */
 public final class EditorSnapshot implements MuSnapshotable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(EditorSnapshot.class);
+
+    @Override
+    public void read(Configuration configuration) {
+        LOGGER.info("Loading snapshot configuration for " + EditorSnapshot.class);
```
pskowronek, line 36: ....... {}

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextViewerSnapshot.java
```diff
@@ -25,12 +31,32 @@
...
 public final class TextViewerSnapshot implements MuSnapshotable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(TextViewerSnapshot.class);
+
+    /**
+     * Section describing information specific to text file presenter.
+     */
+    public static final String TEXT_FILE_PRESENTER_SECTION = FILE_PRESENTER_SECTION + "." + "text";
+
+    @Override
+    public void read(Configuration configuration) {
```
pskowronek, line 42: A lot of dup'd code - why not move it to some base class?

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextViewerPreferences.java
```diff
@@ -78,27 +100,35 @@ enum EditorViewerMode {
...
-        this.value = defaultValue;
-        this.mode = mode;
+
+        Metadata metadata;
+        try {
+            metadata = getDeclaringClass().getDeclaredField(this.name()).getDeclaredAnnotation(Metadata.class);
+            mode = metadata.mode();
+            value = metadata.value();
+        } catch (NoSuchFieldException | SecurityException e) {
+            // ignore, the properties are initialized with the annotation's default values
```
pskowronek, line 130: LOGGER.trace ?

pskowronek (commented):
pls see my comments

ahadas:
thanks @pskowronek, please take another look

pskowronek (approved):
looks good

Issue #896, opened 2023-03-27
Fix nightly builds


aashipov:
create-release fails with
Unable to resolve action `dev-drprasad/delete-tag-and-release@v0.2.0`, unable to find version `v0.2.0`

Source of error - improvements in dev-drprasad/delete-tag-and-release

PR #897, opened 2023-03-27
#896 Fix nightly builds

aashipov:


.github/workflows/nightly.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
     steps:
       - name: Remove previous tag and release
-        uses: dev-drprasad/delete-tag-and-release@v0.2.0
+        uses: dev-drprasad/delete-tag-and-release@v0.2.1
```
ahadas, line 20: @aashipov the previous attempt to upgrade this resulted in losing all the published releases and therefore the change [was reverted](https://github.com/mucommander/mucommander/commit/8f260c4a3b6de045583076bda347c1e87f166825) - need to see that issue is really fixed now

ahadas: or at least that we have a backup of all the released artifacts before this upgrade

ahadas: took backup

ahadas (approved):

Issue #879, opened 2023-02-27
Search options in new editor


Andreas0602:
Suggestion for future improvement of new text editor: Search options, for example like in the internal editor of Double Commander.
<img width="497" alt="Search Options in Double Commander" src="https://user-images.githubusercontent.com/79788982/221508112-c6fe87b1-1503-4e66-b2ab-f6b43e3e78e9.png">

ahadas:
Version 1.3.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.3.0-1)

PR #902, opened 2023-04-13
Search options in new editor

pskowronek:
initial commit for review, will continue to add Replace - #879


pskowronek:
I don't know why `jre` stuck to my commit (due to `rebase` somehow).

mucommander-core/src/main/java/com/mucommander/search/SearchBuilder.java
```diff
@@ -93,19 +93,11 @@ public SearchBuilder where(AbstractFile entrypoint) {
         return this;
     }
 
-    public SearchBuilder depth(int searchDepth) {
+    public SearchBuilder depth(int searchDepth) { // TODO not used?
```
ahadas, line 96: yeah, seems it was replaced with `searchDepth`

mucommander-core/src/main/java/com/mucommander/search/LastSearchQuery.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.mucommander.search;
+
+/**
+ * A class to keep the last search query and its options.
+ */
+public final class LastSearchQuery {
```
ahadas, line 22: why do we need this class, can't we use the values from SearchProperty instead?

mucommander-core/src/main/java/com/mucommander/search/SearchDialog.java
```diff
@@ -241,6 +238,7 @@ private void cleanError() {
         mainPanel.addSpace(10);
 
         YBoxPanel textSearchPanel = new YBoxPanel(10);
+        // TODO FIXME "Text search (Optional)" - not in dictionary?
```
ahadas, line 241: yeah, should be in the dictionary

jre
ahadas, line 1: is this change intentional?

mucommander-translator/src/main/resources/dictionary.properties
```diff
@@ -601,6 +601,13 @@ text_viewer.tabs_emulated = Tabs emulated with spaces
 text_viewer.tab_lines = Tab lines
 text_viewer.tab_size = Tab size
 text_viewer.whitespace_visible = Whitespace visible
+text_viewer.find_replace.backward = Backward
+text_viewer.find_replace.case_sensitive = Case sensitive
```
ahadas, line 605: we can point to another occurrence of 'case sensitive'

mucommander-translator/src/main/resources/dictionary.properties
```diff
@@ -601,6 +601,13 @@ text_viewer.tabs_emulated = Tabs emulated with spaces
...
+text_viewer.find_replace.direction = Direction
+text_viewer.find_replace.forward = Forward
+text_viewer.find_replace.options = Options
+text_viewer.find_replace.regex_match = Regular expressions
```
ahadas, line 609: also here, better to point to an existing entry

jre
ahadas: ah ok, I see your comment about that

mucommander-viewer-text/build.gradle
```diff
@@ -23,7 +23,7 @@ jar {
         'Bundle-Description': 'Library for textual viewer/editor',
         'Bundle-DocURL': 'https://www.mucommander.com',
         'Export-Package': 'com.mucommander.viewer.text',
-        'Import-Package': 'org.fife.ui.rsyntaxtextarea;resolution:=dynamic,*',
+        'Import-Package': 'com.mucommander.ui.text;resolution:=dynamic,com.mucommander.search;resolution:=dynamic,org.fife.ui.rsyntaxtextarea;resolution:=dynamic,*',
```
ahadas, line 26: why is this needed when the packages are exported?

mucommander-core/src/main/java/com/mucommander/search/LastSearchQuery.java
ahadas: posted #903 with the alternative I had in mind

mucommander-viewer-text/build.gradle
ahadas: also covered in #903 (I meant the `resolution:=dynamic` part)

mucommander-core/src/main/java/com/mucommander/search/LastSearchQuery.java
pskowronek: Not sure if I like representing `boolean` values as strings.... but OK, I will follow your idea.

pskowronek:
PR updated with all remarks addressed I think. If it is OK then it can be merged, then I will continue with Replace functionality.

@ahadas any hint how to know inside `TextEditor` if file was opened from search results? I could then run `Find` to move the first occurrence of searched string.

mucommander-translator/src/main/resources/dictionary_en.properties
```diff
@@ -484,7 +484,8 @@ search_dialog.max_threads = Maximum
 search_dialog.case_sensitive = $[file_selection_dialog.case_sensitive]
 search_dialog.matches_regexp = $[file_selection_dialog.matches_regexp]
 search_dialog.wildcards = (* = any string, ? = any character, \\ = escape for literals: *?)
-search_dialog.search_text  = Containing text
+search_dialog.search_text = Containing text
+search_dialog.search_text_optional = Text search (Optional)
```
ahadas, line 488: why here and not in dictionary.properties?

ahadas (approved):

ahadas:
just one minor comment above

> @ahadas any hint how to know inside `TextEditor` if file was opened from search results? I could then run `Find` to move to the first occurrence of searched string.

that would be nice indeed. I don't see a "clean" way to do this but to take the location that is presented in the active folder panel and check if it starts with a search-scheme

mucommander-translator/src/main/resources/dictionary_en.properties
pskowronek: right, I've updated PR

ahadas (approved):

PR #906, opened 2023-04-17
viewer/editor open file from Search and find the first occurrence #879

pskowronek:
If opened file is from Search with Content then scroll (find) to the first occurrence #879


mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
```diff
@@ -1,111 +1,118 @@
...
+    // -----------------------------------------------------------------------------------
+    /**
+     * Edits the currently selected file.
+     */
+    @Override
+    public synchronized void performAction() {
+        AbstractFile file;
+        Command      customCommand;
+
+        boolean inSearch = mainFrame.getActivePanel().getCurrentFolder().getPath().startsWith(
```
pskowronek, line 69: sorry, but I had to convert it to linux EOLs - here starts the change up to line 72.

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -468,7 +467,14 @@ void write(Writer writer) throws IOException {
 
     public void setFocusAndCursorOnFirstLine() {
         textArea.requestFocusInWindow();
-        textArea.setCaretPosition(0);
+
+        // Check if opened file is from Search with content if so, then try to show the found string,
+        // otherwise set caret on the first line.
```
ahadas, line 472: so we need to rename this method

mucommander-core/src/main/java/com/mucommander/search/SearchProperty.java
```diff
@@ -43,6 +43,8 @@ public enum SearchProperty {
     SEARCH_TEXT("text", "search_dialog.search_text", ""),
     TEXT_WHOLE_WORDS("whole-words", "used-via-find-dialog1", Boolean.toString(false)),
     TEXT_SEARCH_FORWARD("forward", "used-via-find-dialog2", Boolean.toString(true)),
+    OPEN_WITH_FOUND_SELECTED(null,"not-used-in-ui", Boolean.toString(false)),
```
ahadas, line 46: I don't think that this is the best place to store it - it's not really a search property, is it?
did you try to inject it all the way to TextEditorImpl?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/AbstractViewerAction.java
```diff
@@ -1,111 +1,118 @@
...
+                SearchFile.SCHEMA + "://");
```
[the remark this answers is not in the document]
pskowronek: yes, it's better

mucommander-core/src/main/java/com/mucommander/search/SearchProperty.java
pskowronek: Didn't want to mess 'all the way' to TextEditorImpl - of course I had the same doubts as you, however, yes, it is not search property, but still conveys search related information :)

pskowronek:
PR updated with review comments excluding the last remark - will think about it.

mucommander-core/src/main/java/com/mucommander/search/SearchProperty.java
pskowronek: PR updated, @ahadas pls check if it is OK now.

pskowronek:
PR updated to address  'all the way' to TextEditorImpl

mucommander-core/src/main/java/com/mucommander/ui/viewer/FileEditorPresenter.java
```diff
@@ -46,6 +46,8 @@
 import com.mucommander.viewer.FileEditor;
 import com.mucommander.viewer.FileEditorService;
 
+import org.slf4j.LoggerFactory;
```
ahadas, line 49: yeah, I've also noticed that we're inconsistent with the logger we use - but that's unused now, right?

package/readme.txt
```diff
@@ -37,6 +37,7 @@ New features:
 
 Improvements:
 - When sorting a file table by filenames, the sort operation performs a locale-sensitive String comparison.
+- More search options to Editor/Viewer, synchronize search options from File Search and set caret on the first occurrence
```
ahadas, line 40: it's better not to cover everything but make it more clear - without listing the search options I don't think it is worth adding the `More search options to Editor/Viewer` part, and without explaining what it means to synchronize the search options I don't think that adding the `synchronize search options from File Search` is worth it.

maybe we can change this to something like: When viewing/editing a file that is returned by file search with text search using the internal viewer/editor, the caret is set on first occurrence of the searched text?

ahadas (commented):
looks good overall, minor comments inside

pskowronek:
PR updated to reflect review comments. BTW, I've moved 'search options' to Features section.

Issue #876, opened 2023-02-26
'Replace' in new editor


Andreas0602:
Suggestion for future improvement of new text editor: Support of Replace functionality including 'replace all', 'replace current', 'find next'.

ahadas:
@pskowronek WDYT?

pskowronek:
I will take a look in #879

pskowronek:
PR provided

ahadas:
Version 1.3.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.3.0-1)

PR #907, opened 2023-04-18
viewer/editor added Find and Replace when in edit mode #879 #876

pskowronek:
Added `Replace` option to Editor via FindDialog (similar to screenshot provided in #879):

obraz

This option does 'replace all' and ignores Forward/Backward settings (they become disabled on UI if checkbox is set). This is due to nature of `RSyntaxTextArea` lib. For Viewer mode this option is not visible.

In Editor mode the menu item `Find` has changed to `Find and replace`


mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/FindDialog.java
```diff
@@ -129,6 +142,18 @@ public FindDialog(JFrame editorFrame) {
 
         groupingPanel.add(directionPanel);
 
+        if (showReplace) {
+            replacePanel.addRow(doReplace, replaceField, 10);
+            compPanel.addRow(replacePanel, 5);
+            doReplace.addItemListener(e -> {
+                boolean selected = e.getStateChange() == ItemEvent.SELECTED;
+                replaceField.setEnabled(selected);
+                // due to nature of RSyntaxTextArea these don't matter for replaceAll operation.
```
ahadas, line 151:
```suggestion
                // due to the nature of RSyntaxTextArea these don't matter for replaceAll operation.
```

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/FindDialog.java
```diff
@@ -163,6 +188,15 @@ public String getSearchString() {
         return findField.getText();
     }
 
+    /**
+     * Returns the reaplce string if 'replace' check box was selected, null otherwise.
```
ahadas, line 192:
```suggestion
     * Returns the replace string if 'replace' check box was selected, null otherwise.
```

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/FindDialog.java
```diff
@@ -163,6 +188,15 @@ public String getSearchString() {
...
+     *
+     * @return the reaplce string if 'replace' check box was selected, null otherwise.
```
ahadas, line 194:
```suggestion
     * @return the replace string if 'replace' check box was selected, null otherwise.
```

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -199,7 +199,8 @@ public void mouseWheelMoved(MouseWheelEvent e) {
     }
 
     void find() {
-        FindDialog findDialog = new FindDialog(frame);
+        boolean isEditable = textArea.isEditable();
```
ahadas, line 202: for consistency, the variable shouldn't start with 'is'
```suggestion
        boolean editable = textArea.isEditable();
```

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -232,23 +234,42 @@ void findPrevious() {
     void convertSpacesToTabs() { textArea.convertSpacesToTabs(); }
 
     private void doSearch(boolean forward) {
+        doSearchAndReplace(forward, null);
+    }
+
+    /**
+     * Does Search or Replace if replaceWith is not null.
```
ahadas, line 241: how about making it clear here that "Replace" means Replace-all?

mucommander-viewer-text/src/main/java/com/mucommander/viewer/text/TextEditorImpl.java
```diff
@@ -232,23 +234,42 @@ void findPrevious() {
...
+     *
+     * @param forward search/replace forward
```
ahadas, line 243: but there's no replace-forward, right? so this need to change to something like
```suggestion
     * @param forward search forward if 'true' or backward otherwise., ignored when replaceWith is not null
```

ahadas (commented):
@pskowronek , nice, looks good overall, minor comments inside

pskowronek:
PR update with review comments

ahadas (approved):
thanks!

Issue #933, opened 2023-05-14
Cannot enter 'pipe' and other symbols in terminal on Mac with German keyboard


Andreas0602:
I cannot enter the 'pipe' symbol `|` in the built-in terminal on a Mac with German keyboard. When I go to the terminal by pressing F12 and then press `Option + 7`, nothing will happen. In Apple's Terminal.app, it does work. Also, it does work in muCommander in the 'Find' dialog or when renaming a file, but it does not work in muCommander's terminal. The following symbols don't work either: `{`, `}`, `[`, `]`. This severely limits the usefulness of muCommander's terminal.

<img width="572" alt="Screenshot_2023-05-14_11-53-25" src="https://github.com/mucommander/mucommander/assets/79788982/c09629b3-d3d3-47fc-8d3a-10e9599789d6">

pskowronek:
Can you try cloning this repo https://github.com/JetBrains/jediterm and run `./jediterm.sh` - are you able to reproduce the problem in their terminal?

pskowronek:
hmm, I've switched input to German and yes, I can reproduce the problem in muC, but also in jeditem master:

![obraz](https://github.com/mucommander/mucommander/assets/360665/801ee1f3-fc22-4a44-b654-64f24fac1e00)

This what happens when I hit `Option + 7` either via keyboard or virtual keyboard on the screen.

pskowronek:
In Idea the had similar problems (umlauts, pipes etc): https://youtrack.jetbrains.com/issue/IDEA-115276 - JDK update apparently fixed the problem there.

pskowronek:
btw, I can this behavior on macOS Terminal when I turn on `Use Option as Meta key` in terminal preferences (see https://osxdaily.com/2013/02/01/use-option-as-meta-key-in-mac-os-x-terminal/)

pskowronek:
`Arg 7` shows only on `bash` imho, on `zsh` is not shown - but the effect is the same (letter multiplied 7 times).

As it comes to my setup - mac laptop with built-in keyboard, right Option works always when it comes to pipe and DE.

pskowronek:
btw, do you use German or German Standard layout?

pskowronek:
![obraz](https://github.com/mucommander/mucommander/assets/360665/9135153f-2950-4bc1-b5b4-07dd2a72dd04)

But I guess it is `German` - only there option-7 should produce pipe

pskowronek:
[...]

PR #935, opened 2023-05-14
Fix for DE keyboard layout #933

pskowronek:
Fix for DE (and some others) keyboard layout: #933

Added preference option (which is by default `false`):

obraz

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalSettingsProvider.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalSettingsProvider.java
@@ -1,0 +1,59 @@
+/*
+ * This file is part of muCommander, http://www.mucommander.com
+ *
+ * muCommander is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * muCommander is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.mucommander.ui.terminal;
+
+import com.jediterm.terminal.ui.settings.DefaultSettingsProvider;
+import com.mucommander.commons.conf.ConfigurationEvent;
+import com.mucommander.commons.conf.ConfigurationListener;
+import com.mucommander.commons.runtime.OsFamily;
+import com.mucommander.conf.MuConfigurations;
+import com.mucommander.conf.MuPreference;
+import com.mucommander.conf.MuPreferences;
+
+/**
+ * A class providing configuration for Terminal - it is based on DefaultSettingsProvider
+ * and overrides certain settings based on Preferences.
+ * Currently, only 'altSendsEscape' driven by USE_OPTION_AS_META_KEY setting (https://github.com/mucommander/mucommander/issues/933)
+ */
+public class TerminalSettingsProvider extends DefaultSettingsProvider implements ConfigurationListener {
+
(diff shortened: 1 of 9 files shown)


mucommander-preferences/src/main/java/com/mucommander/conf/MuPreferences.java
```diff
@@ -232,7 +235,7 @@ public class MuPreferences implements MuPreferencesAPI {
 	/** Default value for 'Quick search timeout' */
 	public static final int DEFAULT_QUICK_SEARCH_TIMEOUT           = 2;
 
-	/** Name of the root element's attribute that contains the version of muCommander used to write the CONFIGURATION file. */
+    /** Name of the root element's attribute that contains the version of muCommander used to write the CONFIGURATION file. */
 	static final String VERSION_ATTRIBUTE = "version";
```
ahadas, line 239: something about the indentation here seems wrong

mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalWindow.java
ahadas, line 1: with this implementation, changes to the configuration won't apply to a terminal that was already initiated, right? I don't think it's that important but why not to register this class as a preferences-listener and track changes of this configuration?

mucommander-preferences/src/main/java/com/mucommander/conf/MuPreferences.java
pskowronek: yup, because this files has tabs as indents. I will make it look ok, won't convert it to spaces tho.

mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalWindow.java
pskowronek: hmm, not sure if it is worth to complicate things related to registering and de-registering listeners.... To have Terminal with new configuration one just need to re-open it. But, if you prefer to have pref listener I can take a look. Let me know.

pskowronek:
PR updated with 1st comment.

mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalWindow.java
ahadas: I don't think you'll need to bother with removing listeners since they are stored using weak references
it would be nicer imo, especially if we'll change the terminal to be presented below the file tables at some point

pskowronek: OK, I will take a look. In the meantime you want to merge it to master as is, I will try to provide another PR later.

pskowronek: PR updated with dynamic preferences, @ahadas

mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalSettingsProvider.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public boolean altSendsEscape() {
+        return altSendsEscape;
+    }
+
+    @Override
+    public void configurationChanged(ConfigurationEvent event) {
+        if (MuPreference.USE_OPTION_AS_META_KEY.equals(
```
ahadas, line 54: how about checking if that's the configuration that has changed without iterating the values of MuPreference by
```suggestion
        if (MuPreferences.USE_OPTION_AS_META_KEY.equals(event.getVariable())) {
```

pskowronek: I don't like to operate on strings if we can use enums.... I would rather say that MuPreferences should be refactored to be enums, too.

ahadas: ok, in my opinion it's alright to use strings now that we can use them in switch-case but that's a matter of taste
I don't mind having an extra step of getting the enum from the label here but we can do it more efficiently - I'll post a separate cleanup

ahadas (approved):
thanks @pskowronek !

Issue #937, opened 2023-05-17
[macOS] New "Open with" feature does not work with File > Open with


Andreas0602:
On macOS, the new "Open with" feature works fine when right-clicking a file and then going to the menu item "Open with ...", but it does not work via the main menu item File > Open with.

pskowronek:
That was by 'design' - see https://github.com/mucommander/mucommander/pull/917 - "Main menu is left as it was".
In that comment there was a question whether we want it in Main Menu or not, and probably how to do it exactly - we don't want to constantly search for apps, probably only when the menu item "open with" is clicked.

pskowronek:
> Obviously you've got two different implementations for creating the menu. I'd suggest using the same implementation for both cases. Just my two cents.

The problem is different. The code for `Open With` generation is the same, parent menus are different tho, and they have different assumptions. Contextual menu is event driven, whereas Main menu is constructed once on creation - there must be introduced contextual knowledge of selected file plus dynamic checking what options are relevant. 

I will try to implement that, but it is not a blocker for release I think.

pskowronek:
@Andreas0602 just pushed PR, can you take a look how it works for you?

PR #966, opened 2023-05-28
Main menu Open With now contains apps on macos

pskowronek:
Main menu Open With now contains apps on macos #937


ahadas (approved):
thanks @pskowronek !

PR #881, opened 2023-02-27
Support multi volume rar files

ShayArtzi:
Hi @ahadas!

I'm creating this PR to discuss some of the challenges in implementing multi volume support. I'm starting with RAR file as the 7z implementation will be different.

Some points I'd like to get your thought on if possible:

1. I created `SevenZipRarMultiVolumeCallbackHandler` to handle multi volume archives and used a regular expression on the file name to detect when it's right to use it. This can be optimized by not compiling the pattern each time but keeping it as a field in the class. What do you think about that approach?
2. I wanted to make sure all volume streams are closed at the end of reading operation, so I created a wrapper class `InArchiveWrapper` that carries over the original `IInArchive` and the handler that references other volume streams.
3. I needed a way to create an `AbstractFile` without recursively calling `SevenZipRarMultiVolumeCallbackHandler::getStream`. I couldn't find a proper way to do it so I changed the modifier of `FileFactory::createRawFile` to `public`. I guess that's not ideal so I'm open to other suggestions (I wanted to reuse the existing `SignatureCheckedRandomAccessFile` class)

Here is a [reference](https://sevenzipjbind.sourceforge.net/extraction_snippets.html#open-multipart-rar-archives) from `sevenzipjbind`'s docs about opening multi volume archives.

Using the code in this PR, I was able to extract a multi-volume RAR archive successfully.

Thanks very much :)

PR #980, opened 2023-06-10
Add support for multi volume 7z files

ShayArtzi:
Hi folks,

I would like to finish the multi-volume archive support I have started in PR #881. This PR is for adding support for multi volume 7z archives.

Multi volume 7z archives have a special file extension - `7z.DDD` where `DDD` is a 3-digit number, starting with `001`:
image

Internally, multi-part 7z files are just a 7z file split into multiple files. It means that:

1. Only the first file starts with a magic number (signature).
2. There is no way to know ahead of time what is the total number of files.

This is different from the way multi volume rar files are implemented, and that's the reason why I have created a dedicated handler for multi volume 7z files (and didn't re-use the multi volume rar handler).

Please let me know what you think... Thanks!


sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
```diff
@@ -0,0 +1,92 @@
+package com.mucommander.sevenzipjbindings.multivolume;
```
pskowronek, line 1: muC copyright header

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                f.close();
+            } catch (Exception e) {
+                LOGGER.error("Error closing IInStream", e);
+            }
+        }
+    }
+
+    @Override
+    public String cryptoGetTextPassword() throws SevenZipException {
+        if (password == null) {
```
pskowronek, line 87: how about empty password? how is it being treated - just being curious....

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                    stream = new SignatureCheckedRandomAccessFile(abstractFile, signature);
+                    isFirstFile = false;
+                } else {
+                    stream = new RandomAccessFileInStream(new RandomAccessFile(filename, "r"));
+                }
+                fileCache.put(filename, stream);
+            }
+            return stream;
+        } catch (FileNotFoundException e) {
+            // There is no way to know ahead of time if we reached the last file,
```
pskowronek, line 66: stil I would add logger.debug here

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public IInStream getStream(String filename) throws SevenZipException {
+        try {
+            IInStream stream = fileCache.get(filename);
+            if (stream != null) {
+                stream.seek(0, ISeekableStream.SEEK_SET);
+            } else {
+                AbstractFile abstractFile = FileFactory.getFile(filename);
```
ahadas, line 54: it's not used when we're not working on the first volume so please inline in the block it's used in

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                } else {
+                    stream = new RandomAccessFileInStream(new RandomAccessFile(filename, "r"));
+                }
+                fileCache.put(filename, stream);
+            }
+            return stream;
+        } catch (FileNotFoundException e) {
+            // There is no way to know ahead of time if we reached the last file,
+            // So it is safe to ignore this Exception
+            return null;
```
ahadas, line 68: what happens when we have only some parts available? can it be that the content of an entry is split across several volumes and so we get corrupted data?
if we have to have all the parts then we need to raise an error in this case (maybe even if the data is not corrupted but only some part of the content can be retrieved)

ahadas (commented):
thanks for the follow-up @ShayArtzi!

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
ShayArtzi: added

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
ShayArtzi: added

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private Map<String, IInStream> fileCache = new HashMap<>();
+
+    private final AbstractFile firstFile;
+
+    private final byte[] signature;
+
+    private final String password;
+
+    private boolean isFirstFile = true;
```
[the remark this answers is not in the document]
ShayArtzi: done

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
ShayArtzi: nice catch - fixed

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
ShayArtzi: If any part is missing, the seven zip library will trigger another exception and the app falls back to the regular error logic. It looks like this:

I renamed the 2nd file to simulate missing volume:
image

And this is what I get when trying to extract:
image

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
ShayArtzi: Interesting question... I don't think it is technically possible to encrypt with an empty password because essentially you don't have a key to perform the encryption operation in the first place...

sevenzipjbindings/src/main/java/com/mucommander/sevenzipjbindings/multivolume/SevenZipMultiVolumeCallbackHandler.java
ahadas: Ack, that's good

ahadas (approved):

ahadas:
thanks @ShayArtzi !

PR #986, opened 2023-06-23
Cleanup for search/replace functionality in binary editor

ahadas:
1. Rephrase the entry about this new functionality in the readme file
2. Increase reuse of existing entries in the dictionary file


pskowronek (approved):
looks good

hajdam (approved):
Seems ok to me. Not sure why `$[text_viewer.find]` is used somewhere instead of `$[find]` or `$[binary_viewer.find]`, but it doesn't matter, I guess...

ahadas:
> Seems ok to me. Not sure why `$[text_viewer.find]` is used somewhere instead of `$[find]` or `$[binary_viewer.find]`, but it doesn't matter, I guess...

yeah, there's a room for further cleanup there - just need to see that other translation files include `find` rather than `text_viewer.find`

Issue #850, opened 2023-01-14
nightly dmg is damaged


superbonaci:
macOS keeps saying "muCommander" is damaged and can't be opened. You should move it to the Bin.
Doesn't happen with stable release.

ahadas:
that's a known issue - the nightly build is not signed, see https://github.com/mucommander/mucommander/discussions/613 for how to install that dmg

superbonaci:
The `sudo xattr -cr /Applications/muCommander.app` doesn't work anymore with macOS 13.1. You must go to System Settings - Privacy & Security - "muCommander" was blocked from use because it is not from an identified developer. Open Anyway. Then it works.

superbonaci:
If you had Settings opened just close it and reopen so "Open Anyway" appears.

superbonaci:
Update: you must `sudo xattr -cr /Applications/muCommander.app`, until you do this you can't unblock it from Systems Settings because doesn't appear.

 Then go to System Settings - Privacy & Security - "muCommander" was blocked from use because it is not from an identified developer. Open Anyway.

Then it works.

georgjaehnig:
@superbonaci When trying `sudo xattr -cr /Applications/muCommander.app`, I get many lines with

    [...]
    xattr: [Errno 1] Operation not permitted: '/Applications/muCommander.app/Contents/PkgInfo'
    xattr: [Errno 1] Operation not permitted: '/Applications/muCommander.app/Contents'
    xattr: [Errno 1] Operation not permitted: '/Applications/muCommander.app'

Also tried `chflags -R nouchg /Applications/muCommander.app` from [here](https://superuser.com/a/40754).

pskowronek:
@georgjaehnig 
Try this `` as described here: https://stackoverflow.com/q/64842819/1715521

If that doesn't help, try this: https://stackoverflow.com/a/64985297/1715521

Let us know which one helped.

georgjaehnig:
@pskowronek So I tried:

    $ sudo xattr -rd com.apple.quarantine /Applications/muCommander.app
    ...
    xattr: [Errno 1] Operation not permitted: '/Applications/muCommander.app/Contents'
    xattr: [Errno 1] Operation not permitted: '/Applications/muCommander.app'

Then:

    $ codesign --force --deep --sign - /Applications/muCommander.app
    /Applications/muCommander.app: replacing existing signature
    /Applications/muCommander.app: Operation not permitted
[...]

PR #1004, opened 2023-07-15
Document post-installation of unsigned dmg

ahadas:


pskowronek:
@ahadas Maybe this information could be added to dmg "installer"?  
It should then mentioned this: https://github.com/mucommander/mucommander/issues/850#issuecomment-1637056377

PR #1013, opened 2023-07-24
gcs: cleanup

ahadas:
Organize imports with the code formatter
Rephrase the entry that was added to the readme.txt file


ahadas:
CC @spakm

PR #1028, opened 2023-08-05
Updated bug_report.yml to encourage ppl for searching similar bug

pskowronek:
Updated bug_report.yml to encourage ppl for searching similar bugs (like for crashes on Ventura :) ).

I haven't tested this change, however I followed github docs regarding `checkboxes` ([see](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)).


ahadas:
> (like for crashes on Ventura :) ).

hehe, yeah, fortunately 999 is a number that is easy to remember :)

ahadas (approved):

Issue #977, opened 2023-06-08
[Enhancement] Support of tabs in terminal


ghost:
### Description

In the terminal that can be opened by pressing F12, it would be useful to have tabs, so that you can work with more than one terminal in parallel.

pskowronek:
How it should behave? 
Now the F12 Terminal when hidden and re-shown is trying to follow the current directory from active file panel. Should it work only for _active_ tab of Terminal?

ahadas:
I'd still love to see the terminal below the file table. I saw it once in a colleague's machine (he used a different file manager) and the way he used it looked really nice - the terminal was shown below and when it got focus, its size increased (and when it lost focus, its size decreased). but that's not directly related to this issue

I wouldn't expect to see tabs in the terminal itself. the way the tabs work in mucommander is that there's a single table and its content (current location) changes according to the location that is attached to each tab but that doesn't really fit for the terminal because ideally we should preserve the 'history' of each tab - I'd say there should be one terminal tab visible and the terminal content/instance should be swapped when the active tab changes

pskowronek:
> Should the tab in the normal interface automatically follow this change? 
That would be hard to implement I think.

Let's first precise about which Terminal we're talking about: F12 or `Run Command`?
Maybe let's leave F12 as is (or tune it to what @ahadas mentioned with having cmd line on bottom - btw, can you get screenshots/video how it looked like?), and add tabs just to `Run Command` aka `Open Terminal` ?

ahadas:
> (or tune it to what @ahadas mentioned with having cmd line on bottom - btw, can you get screenshots/video how it looked like?)

I don't remember which file manager it was (we're talking about something like 10 years ago..) but the colleague [I mentioned above](https://github.com/mucommander/mucommander/issues/977#issuecomment-1595776139) is @MarSik :) Martin, do you recall which file manager with an integrated terminal that you frequently switched to you used during the Python course in TLV?

@pskowronek it worked in a similar way to what we see in [this video](https://www.youtube.com/watch?v=WuaoaQMIGxI)

pskowronek:
> > (or tune it to what @ahadas mentioned with having cmd line on bottom - btw, can you get screenshots/video how it looked like?)
> 
[...]

PR #1033, opened 2023-08-07
Terminal - reworked presentation to show two-liner of Terminal and toggle to maximization

pskowronek:
Terminal - reworked presentation to show two-liner of Terminal and toggle to maximization.

As discussed here: https://github.com/mucommander/mucommander/issues/977#issuecomment-1656641763

MuC initial state:
obraz

After clicking Up on divider (or on divider, or shortcut):

obraz

After clicking Down on divider (or on divider, or shortcut):

obraz

After clicking one more time Down on divider (or when terminal is `exit`ed)

obraz

One can move/drag divider manually then its position is being remember as 'minimized' state (to have more lines than 2), however it divider reaches 2/3rds in height then it is being assumed as 'maximized' state and it will be ignored when hiding.


pskowronek:
> Another problem: In the state where the terminal and both file panes are shown, switching between left and right pane does not trigger a `cd` command in the terminal.

PR updated to address this one

pskowronek:
> Using FlatLaf Light, in the initial state, the divider doesn't look good. And even worse, clicking Up has no effect.

As it comes to 'look good' - what do you mean exactly, @Andreas0602 ?
It looks the same as the existing horizontal splitter (i.e. vertical bar/divider):
obraz

On my mac Up button works - it is not so precise as for `Mac OS X` theme, but it works if you aim right. Have you restarted muC after changing the theme or not?

pskowronek:
> There's definitely a problem in the initial state. The Up arrow doesn't work in that state.

can you check logs? For me it works, when Terminal has to initialize for the first time, it comes around 1-2s (and I've got kinda old macbook).

pskowronek:
> I mean that the divider is not clearly recognizable. At least that's true for the initial state (after starting muCommander) and in the 'maximized terminal' state.

No idea how to make it more visible or obvious, however, if you hang mouse on that divider a tooltip should show up to give you a hint what it does. If somebody knows that Terminal functionality exists then one will just press F12 or sth. 

We might think if we would like to add an option to preferences to start terminal when muC is started and show it in 2-liner mode.

build.gradle
```diff
@@ -139,7 +139,11 @@ runOsgi {
     if (port) {
         javaArgs += ' -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=' + port
     }
-    javaArgs += ' --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED --add-exports java.desktop/com.apple.eawt=ALL-UNNAMED --add-exports java.desktop/com.apple.laf=ALL-UNNAMED  --add-exports java.desktop/com.apple.eio=ALL-UNNAMED'
+    javaArgs += ' --add-opens java.base/java.io=ALL-UNNAMED \
+            --add-opens java.base/java.net=ALL-UNNAMED \
+            --add-exports java.desktop/com.apple.eawt=ALL-UNNAMED \
+            --add-exports java.desktop/com.apple.laf=ALL-UNNAMED  \
+            --add-exports java.desktop/com.apple.eio=ALL-UNNAMED'
```
pskowronek, line 146: just split into separate lines to make it easier to maintain

build.gradle
```diff
@@ -340,7 +344,20 @@ task macAppImage(dependsOn: 'createBundlesDir', type: Exec) {
...
                 '--runtime-image', 'jre/macOS', \
-                '--java-options', '--add-exports java.desktop/com.apple.laf=ALL-UNNAMED --add-exports java.desktop/com.apple.eio=ALL-UNNAMED --add-exports java.desktop/com.apple.eawt=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED --add-opens java.transaction.xa/javax.transaction.xa=ALL-UNNAMED --add-opens java.management/javax.management=ALL-UNNAMED --add-opens java.rmi/java.rmi=ALL-UNNAMED --add-opens java.security.jgss/org.ietf.jgss=ALL-UNNAMED --add-opens java.sql/java.sql=ALL-UNNAMED --add-opens java.base/sun.net.www.protocol.http=ALL-UNNAMED --add-opens java.base/sun.net.www.protocol.https=ALL-UNNAMED', \
+                '--java-options', '\
+                        --add-exports java.desktop/com.apple.laf=ALL-UNNAMED \
+                        --add-exports java.desktop/com.apple.eio=ALL-UNNAMED \
+                        --add-exports java.desktop/com.apple.eawt=ALL-UNNAMED \
+                        --add-opens java.desktop/javax.swing.plaf.basic=ALL-UNNAMED \
+                        --add-opens java.base/java.io=ALL-UNNAMED \
```
pskowronek, line 352: added this line (with javax.swing.plaf.basic). Should I also add it to other places? Like the one above or in mucommander.sh?

Is there some similar place for other platforms like unix/windows? @ahadas

package/unix/mucommander.sh
```diff
@@ -35,4 +35,9 @@ fi
 cd `dirname "$BASE_FOLDER"`
 
 # Starts mucommander.
-$JAVA --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED --add-exports java.desktop/com.apple.eawt=ALL-UNNAMED --add-exports java.desktop/com.apple.laf=ALL-UNNAMED  --add-exports java.desktop/com.apple.eio=ALL-UNNAMED -Djava.library.path=/usr/local/lib -cp mucommander-@MU_VERSION@.jar com.mucommander.main.muCommander $@
+$JAVA --add-opens java.base/java.io=ALL-UNNAMED \
+      --add-opens java.base/java.net=ALL-UNNAMED \
+      --add-exports java.desktop/com.apple.eawt=ALL-UNNAMED \
+      --add-exports java.desktop/com.apple.laf=ALL-UNNAMED  \
+      --add-exports java.desktop/com.apple.eio=ALL-UNNAMED \
+      -Djava.library.path=/usr/local/lib -cp mucommander-@MU_VERSION@.jar com.mucommander.main.muCommander $@
```
pskowronek, line 43: just split to separate lines to make easier to maintain

build.gradle
ahadas: yeah, it should also be added to `java-options` in `winAppImage` and `linuxAppImage`
`mucommander.sh` is used in tgz and portable packages now and I don't regularly use them - seems like it missed few things.. but anyway, as for this change - better to add it there as well

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -206,17 +209,34 @@ public Insets getInsets() {
...
         // Add buttons that allow to collapse and expand the split pane in both directions
-        splitPane.setOneTouchExpandable(true);
+        foldersSplitPane.setOneTouchExpandable(true);
 
         // Disable all the JSplitPane accessibility shortcuts that are registered by default, as some of them
         // conflict with default mucommander action shortcuts (e.g. F6 and F8) 
-        splitPane.disableAccessibilityShortcuts();
+        foldersSplitPane.disableAccessibilityShortcuts();
 
+        verticalSplitPane = new JSplitPane( JSplitPane.VERTICAL_SPLIT, true) {
```
ahadas, line 221:
```suggestion
        verticalSplitPane = new JSplitPane(JSplitPane.VERTICAL_SPLIT, true) {
```

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -206,17 +209,34 @@ public Insets getInsets() {
...
+            @Override
+            public Insets getInsets() {
+                return new Insets(0, 0, 0, 0);
+            }
+        };
+        verticalSplitPane.setBorder(null);
+        verticalSplitPane.setOneTouchExpandable(true);
+        // TODO?
```
ahadas, line 229: I see the comment above:
```
// Disable all the JSplitPane accessibility shortcuts that are registered by default, as some of them
// conflict with default mucommander action shortcuts (e.g. F6 and F8)
```
which seems to hold to the verticalSplitPane as well

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -206,17 +209,34 @@ public Insets getInsets() {
...
+        //verticalSplitPane.disableAccessibilityShortcuts();
+        verticalSplitPane.setTopComponent(foldersSplitPane);
+        verticalSplitPane.setBottomComponent(null);
+
+        JLayeredPane layeredPane = new JLayeredPane();
+        layeredPane.setLayout(new BorderLayout() );
```
ahadas, line 235:
```suggestion
        layeredPane.setLayout(new BorderLayout());
```

mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalIntegration.java
```diff
@@ -0,0 +1,304 @@
...
+                    mainFrame.setCursor(Cursor.getPredefinedCursor(Cursor.WAIT_CURSOR));
+                    terminal = getTerminal(newCwd);
+                    terminal.getTerminalPanel().addFocusListener(new FocusAdapter() {
+                        public void focusGained(FocusEvent e) {
+                            syncCWD(mainFrame.getActivePanel().getCurrentFolder().getAbsolutePath());
+                        }
+                    });
+                    cwd = newCwd;
+                    terminal.getTerminalPanel().addCustomKeyListener(termCloseKeyHandler());
+                    // TODO do this better? For now 2 lines ~height + 20%
```
ahadas, line 141: TODO for myself - comment after I play with that

ahadas (commented):
I see @Andreas0602 played with it already and couple of issues were sorted out (good!), I'll follow up on this PR after playing with it myself, but code-wise it looks good!

pskowronek:
btw, should we change functionality of this toolbar icon? 
obraz

to show this integrated terminal? 
Should we keep "Run command" i.e. "external" Terminal window functionality?

ahadas:
> btw, should we change functionality of this toolbar icon? obraz
> 
> to show this integrated terminal? 

+1 for changing the functionality of that toolbar icon

> Should we keep "Run command" i.e. "external" Terminal window functionality?

I don't think so - now that we use the same things behind the scenes for both things and the integrated terminal works well, I don't see a reason to keep the 'run command' dialog

mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalIntegration.java
ahadas: after playing with it, the height looks good to me
but I'd suggest to not to minimize/hide the terminal when mucommander starts, it would be better to show it, in its minimal size, so users would notice it
I also miss the option to switch back from forth from the terminal to the file table using the keyboard - there's no shortcut for this, right?

pskowronek: I think that since Terminal uses resources, I'm not sure if it is really welcomed change? We may add an options to Preferences to control that?

pskowronek: > I also miss the option to switch back from forth from the terminal to the file table using the keyboard - there's no shortcut for this, right?

You mean when both are visible you would like to switch active panels? (Otherwise F12)

ahadas: no, not to switch active panels but rather something like the following scenario:
1. switch to the terminal
2. make the terminal a big larger
3. start building something (e.g., `mvn install`)
4. switch back to the folder table and navigate to a sub folder there while keep seeing the terminal and the progress of the build in the terminal below

pskowronek: this is what I meant. For that there's no shortcut - we may think about it outside of this PR I think

pskowronek:
PR updated after PR review

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
pskowronek: done

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
pskowronek: done

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
pskowronek: done

mucommander-translator/src/main/resources/dictionary_en.properties
```diff
@@ -241,7 +241,7 @@ CustomizeCommandBar.label = Customize command bar
 ToggleStatusBar.show = Show status bar
 ToggleStatusBar.hide = Hide status bar
 ToggleShowFoldersFirst.label = Show folders first
-ToggleTerminal.label = Show terminal
+ToggleTerminal.show = Show terminal
```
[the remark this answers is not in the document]
pskowronek: done

pskowronek:
> btw, should we change functionality of this toolbar icon? obraz
> 
> to show this integrated terminal? Should we keep "Run command" i.e. "external" Terminal window functionality?

removed and replaced with integrated Terminal

build.gradle
pskowronek: done

mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalIntegration.java
pskowronek: > I think that since Terminal uses resources, I'm not sure if it is really welcomed change? We may add an option to Preferences to control that?

And I think there are two types of users - geeks that are familiar with terminal and civilians that might be scared by "black window" - those who come from Windows' Total Commander. So, for "geeks" I would just add preferences panel to launch terminal in minimized 2-liner form (by default set to `false`).  Tell me what you think and I will add it.

ahadas: > this is what I meant. For that there's no shortcut - we may think about it outside of this PR I think

ah ok, I thought you meant the two file table panels, didn't realize you also referred to the terminal as a panel

> And I think there are two types of users - geeks that are familiar with terminal and civilians that might be scared by "black window" - those who came from Windows' Total Commander. So, for "geeks" I would just add preferences panel to launch terminal in minimized 2-liner form (by default set to false). Tell me what you think and I will add it.

I think it makes sense to make configurable, I'm not sure if disabling it by default make more sense than enabling it by default - if we hide the terminal by default, it's less visible for users that may be interested in using an integrated terminal. how about showing the terminal by default and add an option, like `customize command bar`, that can be used to set that preference you proposed to `false`?

pskowronek: OK, however `customize command bar` even for me is kinda ambiguous, we need a better wording for that

ahadas: sure, we can find better wording
this PR introduce an improvement even without this part, would you like to include it here or do it separately?

pskowronek: > sure, we can find better wording this PR introduce an improvement even without this part, would you like to include it here or do it separately?

I will provide a separate PR for that. In the meantime, I've rebased this PR and added feature description to readme.

ahadas (approved):

pskowronek:
> @pskowronek and @ahadas, in the initial state (after startup), clicking the Up arrow has no effect, and dragging the divider upwards has no effect either. The logs don't contain an error.

https://github.com/mucommander/mucommander/pull/1043

pskowronek:
> In the initial state (after startup), clicking the Up arrow in #1043 is okay now, but dragging the divider upwards still has no effect.

This is by design since Terminal underneath is not initialized. We may think about it along PR for Preferences and command line visibility.

ahadas:
Version 1.4.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.4.0-1)

PR #1036, opened 2023-08-12
Show keyboard shortcuts as notification popup on MainFrame

pskowronek:
A notification that is being shown when a user clicks an action rather than using keyboard. Very helpful to learn new shortcuts if somebody is more into keyboard. It is being inspired by the following Idea plugin:
https://plugins.jetbrains.com/plugin/7345-presentation-assistant

The notification disappears automatically after 3s, and it is shown only if action was triggered not by the keyboard and has accelerators.

Screenshot:
hints

Animation:
hints

Preferences:
obraz


mucommander-core/src/main/java/com/mucommander/ui/notifier/NotificationPopup.java
```diff
@@ -0,0 +1,191 @@
+/*
```
pskowronek, line 1: This class uses (abuses) JPopupMenu - kinda hacky, to be reviewed carefully.

ahadas:
I like the idea!

mucommander-core/src/main/java/com/mucommander/ui/action/MuAction.java
```diff
@@ -17,16 +17,22 @@
 
 package com.mucommander.ui.action;
```
pskowronek, line 18: whether is a good place for that? I like that it is a single point, but on the other hand MuAction starts to get heavy . Move the new functionality to other class and call it from MuAction?

mucommander-core/src/main/java/com/mucommander/ui/notifier/NotifierProvider.java
```diff
@@ -116,6 +123,25 @@ public static void displayBackgroundNotification(final NotificationType notifica
...
+     * @param mainFrame the main frame
+     * @param bgColor background color (no alpha supported)
+     * @param fgColor foreground color
+     * @param notification the notification text
+     * @param timeout the time-out in ms after which the notification disappears
+     */
+    public static void displayMainFrameNotification(JFrame mainFrame,
+                                                    String notification,
+                                                    Color bgColor, Color fgColor,
+                                                    long timeout) {
```
ahadas, line 139: for formatting here seems wrong

mucommander-core/src/main/java/com/mucommander/ui/action/MuAction.java
```diff
@@ -1,406 +1,454 @@
...
+        // it means that mouse click was used, and we may want to show keyboard hints
+        if (e.getSource() == null || !(e.getSource() instanceof FileTable)) {
+            String keyShortcut = Stream.of(
+                            this.getAccelerator() != null
+                                    ? KeyStrokeUtils.getKeyStrokeDisplayableRepresentation(getAccelerator())
+                                    : null,
+                            this.getAlternateAccelerator() != null
+                                    ? KeyStrokeUtils.getKeyStrokeDisplayableRepresentation(getAlternateAccelerator())
+                                    : null)
+                    .filter(s -> s != null && !s.isEmpty())
```
ahadas, line 430:
```suggestion
                    .filter(s -> !StringUtils.isNullOrEmpty(s))
```

mucommander-core/src/main/java/com/mucommander/ui/action/MuAction.java
```diff
@@ -1,406 +1,454 @@
...
+                    .collect(Collectors.joining(" " + Translator.get("or") + " "));
+            if (!keyShortcut.isBlank()) {
+                NotifierProvider.displayMainFrameNotification(mainFrame,
+                        this.getLabel() + ": " + keyShortcut,
+                        new Color(35, 124, 35), Color.WHITE, 3000L);
```
ahadas, line 435: better to replace those last 3 arguments with constants

mucommander-core/src/main/java/com/mucommander/ui/action/MuAction.java
ahadas: the entire file has changed so it's hard to say.. what changed, just `displayKeyboardNotification` and `actionPerformed`?

pskowronek: yes, however, if you select with shift all commits except the first one (MuAction refactor) they will see the exact changes.

mucommander-core/src/main/java/com/mucommander/ui/notifier/NotificationPopup.java
```diff
@@ -0,0 +1,209 @@
...
+            }
+        };
+        panel.setBorder(BorderFactory.createLineBorder(Color.black, 1));
+
+        labelText = new JLabel();
+        panel.add(labelText);
+        popup.add(panel, BorderLayout.CENTER);
+    }
+
+    private static class NotificationPopupHolder {
```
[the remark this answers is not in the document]
pskowronek: it is a way to create Singleton classes

pskowronek:
PR updated with review comments

mucommander-core/src/main/java/com/mucommander/ui/notifier/NotificationPopup.java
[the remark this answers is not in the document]
ahadas: ok but what's the benefit compared to moving the INSTANCE variable to the NotificationPopup class?

mucommander-core/src/main/java/com/mucommander/ui/notifier/NotifierProvider.java
```diff
@@ -116,6 +123,23 @@ public static void displayBackgroundNotification(final NotificationType notifica
...
+     * If given notification is null or blank it does nothing.
+     *
+     * @param mainFrame the main frame
+     * @param bgColor background color (no alpha supported)
+     * @param fgColor foreground color
+     * @param notification the notification text
+     * @param timeout the time-out in ms after which the notification disappears
+     */
+    public static void displayMainFrameNotification(JFrame mainFrame, String notification,
+                                                    Color bgColor, Color fgColor, long timeout) {
```
ahadas, line 137: is it set this way by the formatter? I'd expect the formatter to place `mainFrame` at line 136 and each of the other arguments in a separate line that starts under the 'static' word

pskowronek: ad 1. It is being lazy initiated when accessed (i.e. when its setting in Prefs is enabled)
ad 2. initial reformat of that piece was wrong, putting args under "(" - now it places them as you say - I don't know what was wrong with Idea and Eclipse formatter back then :) PR is about to be updated in seconds.

Issue #1040, opened 2023-08-13
[Bug] ToolBar and CommandBar shortcut description not updated after change in Preferences


pskowronek:
### Is there an existing issue for this?

- [X] I have searched the existing issues (including the closed ones)

### Description

When changing shortcut for an action that is either present in ToolBar or CommandBar their description is not updated in the UI (the shortcut works, so it is only visual aspect).

### Steps to reproduce

- Go to Preferences and change shortcut for Copy action
- Apply the change

### Expected behavior

- the action description should be updated in the UI

### Actual behavior

- In Command Bar copy button still shows Copy [F5] - the same applies to an action from tool bar.

### muCommander version

1.3.0 and the current master

### Java version

doesn't matter

### Operating System type and version

macOS Catalina

ahadas:
Version 1.4.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.4.0-1)

PR #1041, opened 2023-08-13
Fix to update actions in CommandBar and ToolBar if shortcut changed

pskowronek:
Fix to update actions in CommandBar and ToolBar if shortcut changed - #1040

### DIFF
--- a/package/readme.txt
+++ b/package/readme.txt
@@ -47,7 +47,7 @@ Localization:
 Bug fixes:
 - Fixed copying files to Finder.
 - Fixed drag-and-drop of action buttons in the 'Customize command bar' dialog.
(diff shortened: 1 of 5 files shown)


package/readme.txt
```diff
@@ -46,6 +46,7 @@ Localization:
 Bug fixes:
 - Fixed copying files to Finder.
 - Fixed drag-and-drop of action buttons in the 'Customize command bar' dialog.
+- Fixed actions not being updated when their shortcuts changed in Preferences.
```
ahadas, line 49: how about being specific here that it's about actions in the command bar and toolbar?

ahadas (commented):
looks good, minor comments

pskowronek:
PR updated after review

ahadas (approved):

PR #1043, opened 2023-08-15
Terminal - fix command line #1033

pskowronek:
Terminal - fix command line as mentioned here: [#1033, ](https://github.com/mucommander/mucommander/pull/1033#issuecomment-1678906745)

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
@@ -81,7 +81,7 @@ public class ToggleTerminalAction extends ActiveTabAction {
     }
 
     private void ensureIntegrationIsInitialized() {
-        if (terminalIntegration == null && mainFrame != null && mainFrame.getVerticalSplitPane() != null) {
+        if (terminalIntegration == null && mainFrame.getVerticalSplitPane() != null) {
             // doing it lazy, because in c-tor main frame might not be fully built (no vertical split pane yet)
             synchronized (this) {
                 if (terminalIntegration == null) {
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
@@ -45,14 +45,7 @@ public class ToggleTerminalAction extends ActiveTabAction {
 
     @Override
     public void performAction() {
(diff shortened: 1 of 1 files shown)


pskowronek:
@ahadas a fix - unless you know a better way

ahadas (commented):
@pskowronek I'm not sure that I understand the problem that is fixed by that - where do we use `terminalIntegration` in a flow that calls `toggleEnabledState` without performing the action?

pskowronek:
> `terminalIntegration`

only in Main Frame, it just must integrate with it. Now it is being done in lazy way (when action is clicked that it created Terminal and add it to MainFrame, but since we wanted to also react for clicks on up/down/split then it must add it self to them).

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
```diff
@@ -82,8 +75,19 @@ public String getLabel() {
 
     @Override
     protected void toggleEnabledState() {
+        ensureIntegrationIsInitialized();
         AbstractFile currentFolder = mainFrame.getActivePanel().getCurrentFolder();
         setEnabled(currentFolder.getURL().getScheme().equals(LocalFile.SCHEMA));
     }
 
+    private void ensureIntegrationIsInitialized() {
+        if (terminalIntegration == null && mainFrame != null && mainFrame.getVerticalSplitPane() != null) {
```
ahadas, line 84: `mainFrame` cannot be null here because if it is null then we'll get NPE at line 79
and I'm not sure about the last part (`mainFrame.getVerticalSplitPane() != null`) - but maybe the right place to call `new TerminalIntegration(mainFrame, mainFrame.getVerticalSplitPane()` would be after initializing the vertical split pane rather than this class?

pskowronek: I didn't want to glue vertical split pane to terminal functionality. It is just a container that can be used by other features in the future. I'm about to remove `mainFrame != null` - good point. PR will be updated soon.

> would be after initializing the vertical split pane rather than this class?
I can think about it this place is not good enough. Just let me know.

ahadas: > I can think about it if this place is not good enough. Just let me know.

I find it not intuitive to call it from toggleEnabledState, I'd prefer to call it explicitly from where we expect it the integration to be initialized but your explanation makes sense so let's get this in and continue to think about separately

Issue #492, opened 2021-06-06
muCommander startup was faster with OSX 10.11 and muCommander 0.9.0


bambi24x:
I'm currently using muCommander 0.9.7-1 on macOS 11.3.1. Would be great if the startup speed could be improved.

Thank you very much!

GJRobert:
Indeed. I concur.

pskowronek:
@bambi24x & @GJRobert 
Was Bonjour service discovery enabled? If so, have you tried to turn it off?

GJRobert:
> @bambi24x & @GJRobert Was Bonjour service discovery enabled? If so, have you tried to turn it off?

Hi @pskowronek I have not disabled Bonjour discovery before. I just tried disabling it in muCommander, but it still takes more than 10 sec. to start on my macOS 11 + M1 MacBook Air.

Regards

pskowronek:
Can you re-try with the newest muC version?

GJRobert:
Hi @pskowronek,
I have done a reinstallation of muC 1.2.0 (simple overwriting installation over old version won't launch),
(have to clean old muC 0.9.x first)
and the startup time takes more than 10 sec (a splashscreen does appear earlier notifying about initializing window).
(Bonjour discover turned off as before)

I appreciate your work~~~

pskowronek:
I would just optimize start up first. 

@Andreas0602 - are you able to record a movie from starting muC up until main window is shown? 

If you can, can you also run it directly from sources (./gradlew clean run) and attach console logs as a zip file? We could identify the bottleneck.

On my quite old macbook muC starts at 6-7s.

pskowronek:
Thanks. 
If you disable Bonjour in Prefs does it change anything?

pskowronek:
> Disabling Bonjour doesn't change anything. BTW, if I remember correctly, there was a recent change so that muCommander does not wait for Bonjour during startup.

yup, but in your start up logs I see that some initialization is being done (it takes around 200ms).
`2023-08-13 18:17:05.725 [FelixStartLevel] INFO  c.m.commons.runtime.OsFamily:70 - Current OS family: macOS` - here Felix (osgi) starts to live
`2023-08-13 18:17:06.067 [0.local.).Timer] DEBUG javax.jmdns.impl.DNSIncoming:210` - some Bonjour witchery (lots of entries)
[...]

PR #1044, opened 2023-08-15
muC faster start up

pskowronek:
- tweaked how bundles are being loaded (added parallelized bundle loading and deferring) - this gave 2-3 times faster bundle loading etc (from 2-3s to ~400-800ms)
- added java params for to start faster: `-XX:+TieredCompilation -XX:TieredStopAtLevel=1` - these gave quite a lot of boost (https://stackoverflow.com/a/46516509/1715521)
- further defer Bonjour loading (let JmDNSImpl start later to let muC have resources for start up)
- muCommander.java - parallelized loading of configurations
- replaced all imports swing.* and awt.* in java - this gave almost no speed effect
- SMB activator moved to background (sb needs to check if it works - this never worked for me)
- theme loader - replaced all ladder IFs with switches (hence code compatibility set to 20) - done automagically by Idea (refactor)
- enabling Notifications moved to background (as it also took some time)
- loading icons used .getResource two times
- LaFs loading was taking long - additional LaFs are being loaded in background (unless they are chosen by the user)
- Clipboard check is done on start up, sometimes it caused delay if something unsupported by muC was there (like code copied from Idea, this caused  'class not found' exceptions) - to be addressed here: https://github.com/mucommander/mucommander/issues/1137
- ADB menu deferred, Bonjour init also moved after start
- Quick list loaded in background 
- Pre-load JFrame and couple of JPanels as early as possible and then use it Application for MainFrame etc
- Pre-load all fonts into JVM - this is very slow, at least on macs - start loading fonts in background even before felix is initiated (this may yield caching fonts done in a previous commit unnecessary)

Those changes make loading times 2-3 faster (3s->1s on Andreas' mac, on my mac from 6-7s->~2.6-3s).

Remarks:
- it seems that splash screen loading adds 100-200ms to startup time on my mac (this can be turned off in Preferences)
- What's left is:
 - the time of loading/initializing JVM - not sure if something could be done about it more (@ahadas )
 - think about loding bundles for extensions/plugins when main app started (akin how Idea or Eclipse does it).


mucommander-protocol-adb/src/main/java/com/mucommander/commons/file/protocol/adb/AndroidMenu.java
```diff
@@ -98,7 +98,7 @@ public void menuSelected(MenuEvent e) {
         // Remove previous menu items (if any)
         removeAll();
 
-        List<String> androidDevices = AdbUtils.getDevices();
+        List<String> androidDevices = null; //AdbUtils.getDevices(); - defer loading like for Bonjour
```
pskowronek, line 101: @ahadas defer it?

ahadas:
> btw, I've spotted that `hamcrest-core` bundle is being loaded? Why?

junit drags it:
```
+--- project :mucommander-protocol-registry
|    +--- project :mucommander-commons-file (*)
|    +--- project :mucommander-commons-util (*)
|    +--- project :mucommander-protocol-api (*)
|    +--- project :mucommander-format-gzip (*)
|    +--- project :mucommander-process (*)
|    +--- project :mucommander-translator (*)
|    +--- org.slf4j:slf4j-api:1.7.36
|    +--- org.osgi:osgi.core:8.0.0
|    +--- org.ovirt.engine.api:sdk:4.4.5
|    |    +--- org.slf4j:slf4j-api:1.7.7 -> 1.7.36
|    |    +--- com.fasterxml.jackson.core:jackson-databind:2.10.0
|    |    |    +--- com.fasterxml.jackson.core:jackson-annotations:2.10.0
|    |    |    \--- com.fasterxml.jackson.core:jackson-core:2.10.0
|    |    +--- org.apache.httpcomponents:httpclient:4.5 (*)
|    |    \--- org.apache.httpcomponents:httpcore:4.4.1
|    \--- com.googlecode.json-simple:json-simple:1.1.1
|         \--- junit:junit:4.10
|              \--- org.hamcrest:hamcrest-core:1.1
```

mucommander-protocol-adb/src/main/java/com/mucommander/commons/file/protocol/adb/AndroidMenu.java
ahadas: sure, it doesn't worth the time on startup

pskowronek:
> > btw, I've spotted that `hamcrest-core` bundle is being loaded? Why?
> 
> junit drags it:

any hint where to add exclusion?

pskowronek:
PR updated to address: https://github.com/mucommander/mucommander/issues/492#issuecomment-1679315040

pskowronek:
PR updated - faster UI start up.

pskowronek:
@ahadas - this PR is more PoC, but - in one of the recent commits I've switched code compatibility to java 20 - I guess this broke tests here? `./gradlew clean test` locally works for me.

pskowronek:
@ahadas  - I tried to commit each attempt/idea in a separate commit. Some ideas were removed in subsequent commits (try & error), but overall I think all commits bring some benefit. Font caching is disputable, however I would keep it (it seems that on my mac it saves ~100ms if/when fonts are pre-loaded on time). Since a lot of concurrency has been introduced it requires a lot of testing on different platforms. From my side, occasionally, 1 time I've spotted a problem when action menu bar was not displayed (right now I cannot reproduce that, but I think there's a race condition), plus rarely  I get `Cannot invoke "javax.swing.JButton.setToolTipText(String)" because "oneTouchButton" is null`. 

Nevertheless, tell me what you think about it, should be continue this, or not.

ahadas:
> @ahadas - I tried to commit each attempt/idea in a separate commit. Some ideas were removed in subsequent commits (try & error), but overall I think all commits bring some benefit. Font caching is disputable, however I would keep it (it seems that on my mac it saves ~100ms if/when fonts are pre-loaded on time). Since a lot of concurrency has been introduced it requires a lot of testing on different platforms. From my side, occasionally, 1 time I've spotted a problem when action menu bar was not displayed (right now I cannot reproduce that, but I think there's a race condition), plus rarely I get `Cannot invoke "javax.swing.JButton.setToolTipText(String)" because "oneTouchButton" is null`.
> 
> Nevertheless, tell me what you think about it, should be continue this, or not.

@pskowronek thanks for the effort you've put on this! I was not able to follow the changes closely recently but I read your conversation with @Andreas0602 and it seems you've managed to accelerate the standup time nicely. there are quite a few changes here so it can take me some time to review them but as I want to release 1.4.0 soon, I'll try to provide feedback shortly

src/main/java/com/mucommander/main/AutoProcessor.java
```diff
@@ -74,9 +89,17 @@ public class AutoProcessor
     **/
     public static void process(Map configMap, BundleContext context)
     {
-        configMap = (configMap == null) ? new HashMap() : configMap;
-        processAutoDeploy(configMap, context);
-        processAutoProperties(configMap, context);
+        ExecutorService executor = null;
+        try {
+            executor = Executors.newFixedThreadPool(8);
```
ahadas, line 94: does it need to be within the try-finally block?..

src/main/java/com/mucommander/main/AutoProcessor.java
```diff
@@ -141,52 +164,62 @@ private static void processAutoDeploy(Map configMap, BundleContext context)
             // Look in the specified bundle directory to create a list
             // of all JAR files to install.
             File[] files = new File(autoDir).listFiles();
-            List jarList = new ArrayList();
+            List<File> jarList = new ArrayList<>();
             if (files != null)
             {
                 Arrays.sort(files);
-                for (int i = 0; i < files.length; i++)
+                for (File f : files)
```
ahadas, line 171:
```suggestion
                for (File file : files)
```

src/main/java/com/mucommander/main/AutoProcessor.java
```diff
@@ -141,52 +164,62 @@ private static void processAutoDeploy(Map configMap, BundleContext context)
...
                     // If the bundle is not already installed, then install it
                     // if the 'install' action is present.
                     if ((b == null) && actionList.contains(AUTO_DEPLOY_INSTALL_VALUE))
                     {
-                        b = context.installBundle(
-                            ((File) jarList.get(i)).toURI().toString());
+                        var jarPath = jarFile.toURI().toString();
+                        if (jarPath.contains("dropbox") || jarPath.contains("jediterm")
+                                || jarPath.contains("protocol-gcs") || jarPath.contains("viewer-pdf")
+                                || jarPath.contains("gdrive")) {
```
ahadas, line 200: that's not that great.. what's the reason for this?

src/main/java/com/mucommander/main/AutoProcessor.java
```diff
@@ -219,20 +255,36 @@ else if ((b != null) && actionList.contains(AUTO_DEPLOY_UPDATE_VALUE))
...
+                        System.out.println("------ get bundle " + (System.currentTimeMillis() - pre) + "ms " + bundle);
+                        var startLvl = startLevel;
+                        executor.execute(() -> {
+                            try
+                            {
+                                if (bundle != null) {
+                                    if (!isFragment(bundle)) {
+                                        sl.setBundleStartLevel(bundle, startLvl);
+                                    }
+                                    bundle.start();
```
ahadas, line 277: previously, we wouldn't have started a bundle for which `isFragment(bundle) == true`, did you intentionally change this?

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas, line 1: how about posting the changes here to [apache-felix](https://github.com/apache/felix-dev/blob/master/main/src/main/java/org/apache/felix/main/AutoProcessor.java), it's better as: (1) they can be used by other applications; (2) we'll get feedback from the developers of apache-felix on these changes; and (3) once the changes get it on apache-felix, it will be easier for us to rebase on newer version of apache-felix

mucommander-bonjour/src/main/java/com/mucommander/bonjour/BonjourDirectory.java
ahadas, line 1: the changes related to Bonjour look good

mucommander-core/src/main/java/com/mucommander/ui/main/SplashScreen.java
```diff
@@ -21,16 +21,27 @@
 import com.mucommander.commons.util.ui.dialog.DialogToolkit;
 import com.mucommander.ui.icon.IconManager;
 
-import javax.swing.*;
-import java.awt.*;
-
+// if we want SplashScreen to load faster,
+// then don't load everything from awt.* or swing.* (i.e. no wildcards)
```
ahadas, line 25: no need for this comment - we should avoid imports with wildcards anyway as our formatter doesn't accept it

ahadas:
@pskowronek that's a very partial review - I'm also in the process of trying to extract unrelated changes to make it easier to review

build.gradle
```diff
@@ -40,7 +40,7 @@ repositories.mavenCentral()
 subprojects {
     apply plugin: 'biz.aQute.bnd.builder'
     apply plugin: 'java-library'
-    compileJava.options.compilerArgs += ['--release', '11']
+    compileJava.options.compilerArgs += ['--release', '20']
```
ahadas, line 43: what's the reason for it? unless there's a good reason to bump the code compatibility, I'd like to require the minimal version that's still supported so it would be easier for others to use the code also on platforms that don't have latest JRE

mucommander-core/src/main/java/com/mucommander/Application.java
```diff
@@ -451,52 +499,75 @@ private void run() {
...
-                            MuPreferences.DEFAULT_ENABLE_SYSTEM_NOTIFICATIONS)) {
-                printStartupMessage("Enabling system notifications...");
-                if (com.mucommander.ui.notifier.NotifierProvider.isAvailable())
-                    com.mucommander.ui.notifier.NotifierProvider.getNotifier().setEnabled(true);
-            }
+                // Dispose splash screen.
+                var localSplashScreen = splashScreen;
+                if (localSplashScreen != null) {
+                    localSplashScreen.dispose();
+                    splashScreen = null;
```
ahadas, line 530: can we make `splashScreen` a local variable instead?

mucommander-core/src/main/java/com/mucommander/ui/main/SplashScreen.java
ahadas, line 1: please revert the changes here

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
ahadas, line 1: can't we place this class in mucommander-commons-util?

mucommander-core/src/main/java/com/mucommander/ui/main/SplashScreen.java
ahadas: done in https://github.com/mucommander/mucommander/pull/1057

mucommander-translator/src/main/resources/dictionary.properties
```diff
@@ -547,8 +547,9 @@ server_connect_dialog.dropbox.load_account = Load account
 server_connect_dialog.dropbox.username = $[server_connect_dialog.username]
 server_connect_dialog.dropbox.account = $[server_connect_dialog.account_alias]
 adb.android_devices = Android
-adb.no_devices = No device discovered
 adb.android_disabled = Android disabled
+adb.no_devices = No device discovered
+adb.searching = Searching...
```
ahadas, line 552: we already have "Loading..." translated, how about using it instead?

mucommander-bonjour/src/main/java/com/mucommander/bonjour/BonjourDirectory.java
ahadas: @pskowronek how about separating these changes our to a separate PR?

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: I would prefer to have it in finally, if sth goes wrong we would like to finish jobs in executor before going anymore further. Yeah, I know that muC will be in strange condition, but at least more tasks will be finished and maybe a better log will be available.

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: yes, it's not, I agree. Those bundles are relatively big and not required as the first ones. We may introduce sth like checking bundle size, and defer the biggest ones.

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: I guess my mistake. I will move .start it into if-clause

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: Interesting, didn't know it comes from apache-felix. I will take a look.

build.gradle
pskowronek: It was for theme loader refactor to replace if-else with switch/case with arrows. This was introduced in java 14, I set 20 cause I thought that 20 is our target right now.

mucommander-core/src/main/java/com/mucommander/Application.java
pskowronek: I guess not, it is being used by Application#printStartupMessage

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
pskowronek: I will see on that after rebase.

mucommander-translator/src/main/resources/dictionary.properties
pskowronek: I thought that 'searching' better describes what it does - I can use 'loading', but 'searching' is better for me. If you insist I can change that to 'loading'.

build.gradle
pskowronek: set to 14 in PR update

pskowronek:
PR updated with review comments (and rebased)

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas: ok, but even in a scenario that is unlikely to happen in which we get runtime exception from the call to newFixedThreadPool, there won't be job submitted yet, right?

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas: I'm ok with using several threads to load the bundles but we need to ensure that the protocol and format bundles are loaded before mucommander-core is loaded as we may have an initial location that requires them

build.gradle
ahadas: right, we bundle Java 20 in the rpm, deb, msi and dmg packages but keep the oldest Java version for both the other packages we build and for rebuilders, like [this](http://slackbuilds.org/repository/15.0/system/mucommander/). changing to Java 14 is better but it would be even better to keep Java 11 for the time being, we can still have switch-case but without the arrows..

mucommander-core/src/main/java/com/mucommander/Application.java
ahadas: maybe we can inject it to printStartupMessage as it's called from a single method, need to try

mucommander-translator/src/main/resources/dictionary.properties
ahadas: I don't see much difference and considering that we don't get translation updates so often, it's better to avoid adding new ones unless we really need to

build.gradle
pskowronek: PR updated - converted arrow switches to old style switches (using idea), plus reverted to 11

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
pskowronek: I see that `mucommander-commons-util` has dependency on `api project(":mucommander-commons-runtime")`.

What I wanted to achieve is very light bundle w/o deps so it starts first.
However, not sure why, by core-preload is being put into `core` jar somehow.

mucommander-translator/src/main/resources/dictionary.properties
pskowronek: PR updated

mucommander-core/src/main/java/com/mucommander/Application.java
pskowronek: will you try? I mean, who's gonna try :) 
and I'm not sure how to inject that splash instance into printStartupMessage

pskowronek:
@ahadas - another PR update, dirty fix for tests - let's think where to 'initialize' PreloadedJFrame (OsVersion wasn't meant to be the final place :) ).

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: PR in felix: https://github.com/apache/felix-dev/pull/236

mucommander-bonjour/src/main/java/com/mucommander/bonjour/BonjourDirectory.java
pskowronek: separated: https://github.com/mucommander/mucommander/pull/1061

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -137,72 +152,75 @@ private void setWindowIcon() {
...
+            this.toolbarPanel = ToolbarMoreButton.wrapToolBar(toolbar);
+            this.toolbarPanel.setVisible(MuConfigurations.getPreferences().getVariable(MuPreference.TOOLBAR_VISIBLE, MuPreferences.DEFAULT_TOOLBAR_VISIBLE));
+            contentPane.add(toolbarPanel, BorderLayout.NORTH);
+        });
 
-        // Create menu bar (has to be created after toolbar)
-        MainMenuBar menuBar = new MainMenuBar(this);
-        setJMenuBar(menuBar);
+        executor.execute(() -> {
+            // Create menu bar (has to be created after toolbar) - ok, but why?
```
ahadas, line 207: I doubt anyone remembers, that's a comment from 16 years ago..

mucommander-core/src/main/java/com/mucommander/ui/main/FolderPanel.java
```diff
@@ -130,92 +136,101 @@ it calls actionPerformed() each time an item is highlighted with the arrow (UP/D
...
+
+            // Allow the location field to change the current directory when a file/folder is dropped on it
+            FileDropTargetListener dropTargetListener = new FileDropTargetListener(this, true);
+            new DropTarget(locationTextField, dropTargetListener);
+            new DropTarget(driveButton, dropTargetListener);
+            locationTextField.addFocusListener(this);
+        }).start();
+
+        // Initialize quick lists in background
+        fileTablePopups =  CompletableFuture.supplyAsync(() -> {
```
ahadas, line 177: the initialization of the quick lists should be fairly quick as their data is queried before they are shown - what's the impact of this change?

mucommander-core/src/main/java/com/mucommander/ui/main/FolderPanel.java
```diff
@@ -130,92 +136,101 @@ it calls actionPerformed() each time an item is highlighted with the arrow (UP/D
...
+            locationPanel.add(locationTextField, c);
+            disableCtrlFocusTraversalKeys(locationTextField);
+            registerCycleThruFolderPanelAction(locationTextField);
+
+            // Allow the location field to change the current directory when a file/folder is dropped on it
+            FileDropTargetListener dropTargetListener = new FileDropTargetListener(this, true);
+            new DropTarget(locationTextField, dropTargetListener);
+            new DropTarget(driveButton, dropTargetListener);
+            locationTextField.addFocusListener(this);
+        }).start();
```
ahadas, line 174: did you measure the impact of placing this in a thread? I suppose it's insignificant

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
```diff
@@ -247,20 +265,22 @@ public Insets getInsets() {
...
-        this.statusBar = new StatusBar(this);
-        southPanel.add(statusBar);
-
-        // Show command bar only if it hasn't been disabled in the preferences
-        this.commandBar = new CommandBar(this);
-        // Note: CommandBar.setVisible() has to be called no matter if CommandBar is visible or not, in order for it to be properly initialized
-        this.commandBar.setVisible(MuConfigurations.getPreferences().getVariable(MuPreference.COMMAND_BAR_VISIBLE, MuPreferences.DEFAULT_COMMAND_BAR_VISIBLE));
-        southPanel.add(commandBar);
-        insetsPane.add(southPanel, BorderLayout.SOUTH);
+        executor.execute(() -> {
```
ahadas, line 268: I wonder if we really save much by using several threads here and above - did you check how much we get from that?

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
```diff
@@ -0,0 +1,99 @@
...
+
+import javax.swing.JFrame;
+import javax.swing.JPanel;
+
+/**
+ * A class that extends JFrame to be later on used by core.
+ * Since this bundle is loaded almost as the first (as it has no deps), it can be
+ * used by core having JFrame "preloaded" aka "cached" in JVM.
+*/
+public class PreloadedJFrame extends JFrame {
```
ahadas, line 35: I see that you've made quite a few changes to use this class but all it does is saving us from calling few constructors from the mucommander-core bundle? is it really significant?

ahadas (commented):
@pskowronek I have a feeling that by keeping just a few of the posted changes would result in almost the same execution time, so maybe a better approach would be to identify the major boosts, get them in one by one and then see where we stop (e.g., I'd start with the Java arguments that you've noticed that improves things significantly) 

I'm also concerned by loading the file format and protocols in separate threads - if we get into a situation that the application can't start because of (or shows an error for) a location that couldn't be initialized then the reduction in execution time won't worth it. maybe we should get the initial locations before starting mucommander-core and see what can be started in the background?

I didn't go over all the changes yet but I think most of the changes I didn't look at are related to the replacement of the inheritance with decorator in MainFrame and FolderPanel

build.gradle
```diff
@@ -136,9 +136,11 @@ createBundlesDir.doLast {
 
 runOsgi {
     javaArgs = '-Djava.library.path=/usr/local/lib' // for Libguestfs
+    javaArgs += ' -Xshare:auto -XX:-UsePerfData -XX:+TieredCompilation -XX:TieredStopAtLevel=1' // faster start
+
     def port = System.getenv('DEBUG')
     if (port) {
-        javaArgs += ' -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=' + port
+        javaArgs += ' -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=' + port
```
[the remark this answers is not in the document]
pskowronek: it's been committed by mistake, we can leave it as 'n'

mucommander-core/src/main/java/com/mucommander/ui/main/FolderPanel.java
pskowronek: I've seen considerable delay on it - I would keep it

mucommander-core/src/main/java/com/mucommander/ui/main/FolderPanel.java
pskowronek: I don't remember the numbers :/ but showing that drive button took enough time to give my attention into it.

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: I think that `finally` is kinda a good practice that I would keep it.

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
pskowronek: Don't remember the numbers, it was enough to grab my attention :) command bar and status bar have their icons etc, so it adds a delay if we do it in series. On my relatively slow macbook, sometimes I see that main window is presented and status bar is being added a moment later. So having parallel work on it gives that 'responsiveness' feeling.

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
pskowronek: The main thing is that first initialization of any more complicated components of SWT/AWT is painfully slow (really slow), that's why I start building a couple of JFrames/JPanels as early as possible so when all the bundles are up, we just can put components in it, and SWT/AWT libs are also warmed-up already and therefore execute quickly as they should.

Preloading of those JFrames/JPanels (while OSGi/Felix is still trying to load all the staff) gave a lot of boost.

mucommander-commons-file/src/main/java/com/mucommander/commons/file/icon/impl/SwingFileIconProviderImpl.java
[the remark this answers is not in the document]
pskowronek: done, #1069

build.gradle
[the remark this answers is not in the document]
pskowronek: PR #1070 - after reconsideration, maybe let's have it 'y' by default?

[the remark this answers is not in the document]
ahadas: yeah, that makes more sense - merged the PR

mucommander-core/src/main/java/com/mucommander/ui/main/FolderPanel.java
ahadas: the thing is that if it really takes considerable time, maybe it's not as simple construction as it's supposed to be and we should rather initialize them in a lazy way? I don't think quick-lists are used so frequently (and I'm writing this even though I like this feature - I've implemented it :) )

pskowronek: And they are awesome!
Why are you worried about async?

ahadas: > And they are awesome! 

good to see I'm not the only one using them :)

> Why are you worried about async?

it makes the code more complex - need to think whether something is initialized or not..
I don't have a strong objection but I'd prefer to avoid it when it's not really required

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
ahadas: I see but I wonder whether it's a better approach - I understand the need to reduce the time until users can a fully function window but getting "half baked" window more quickly is not necessarily better I think

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
ahadas: > I see that `mucommander-commons-util` has dependency on `api project(":mucommander-commons-runtime")`.
> 
> What I wanted to achieve is very light bundle w/o deps so it starts first. 

I see, ok - we can probably drop that dependency which comes from FocusDialog but the PreloadedJFrame doesn't sound like a utility class that can be reusable, it's value comes from loading it early so never mind

> However, not sure why, but core-preload is being put into `core` jar somehow.

really? that's weird. you didn't configure it as an OSGi bundle (as it's only used as compileOnly) but still I wouldn't expect to see the PreloadedJFrame class in mucommander-core...

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
ahadas: ack
so why not to define it as a proper OSGi bundle with an activator class?

mucommander-core/src/main/java/com/mucommander/Application.java
ahadas: sure, posted https://github.com/mucommander/mucommander/pull/1071

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas: ok

mucommander-core/src/main/java/com/mucommander/Application.java
pskowronek: ok, I get it - I was thinking that you had some really advanced magic on your mind.

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
pskowronek: > really? that's weird. you didn't configure it as an OSGi bundle (as it's only used as compileOnly) but still I wouldn't expect to see the PreloadedJFrame class in mucommander-core...

yup, that's strange

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: I may try to drop this deferring and see how it impacts the load time.

mucommander-core/src/main/java/com/mucommander/Application.java
pskowronek: PR updated

mucommander-core/src/main/java/com/mucommander/ui/main/MainFrame.java
pskowronek: usually the window is fully loaded quicker, sometimes for blink of second you see those menus appear with slight of latency (And you have to remember that I'm using a macbook from 2012 :) ).

mucommander-core-preload/src/main/java/com/mucommander/preload/PreloadedJFrame.java
pskowronek: because it takes around 1-2s to load OSGi framework (felix) and then to load jars bundles. While waiting for Felix to load we can use that time to init SWT/AWT in other thread (even my 10yr macbook has two cores, effectively 4 hyper-thread cores - we can employ them :)).

pskowronek:
@ahadas I've just rebased the code. Any thoughts what next with this PR?

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
```diff
@@ -129,11 +119,39 @@ private WindowManager() {
             LOGGER.debug("Could load look'n feel from preferences");
         
         MuConfigurations.addPreferencesListener(this);
+
+        if (!additionalLafsLoaded) {    // defer loading other LAFs
```
ahadas, line 123: in which scenario would we get here with additionalLafsLoaded==true?

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
```diff
@@ -181,25 +199,27 @@ public static void tryRefreshCurrentFolders() {
     /**
      * Creates a new MainFrame and makes it visible on the screen, on top of any other frames.
      *
-     * @param leftFolders initial paths for the left frame.
-     * @param rightFolders initial paths for the right frame.
+     * @param mainFrameBuilder the mainFrame builder
      * @return the newly created MainFrame.
      */
     public static synchronized void createNewMainFrame(MainFrameBuilder mainFrameBuilder) {
+        LOGGER.error("creating new main frame...");
```
ahadas, line 206: remove the 'error' logs from here as well?

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
```diff
@@ -260,13 +281,25 @@ private void setLookAndFeel(String lnfName) {
...
             currentThread.setContextClassLoader(ExtensionManager.getClassLoader());
 
-            UIManager.setLookAndFeel(lnfName);
+            try {
+                var currentLaF = UIManager.getLookAndFeel();
+                if (currentLaF == null || !lnfName.equals(currentLaF.getClass().getName())) {
+                    UIManager.setLookAndFeel(lnfName);
+                }
+            } catch (UnsupportedLookAndFeelException e) {
+                LOGGER.info("The chosen Look and feel (that is non-standard) must be loaded now...");
```
ahadas, line 290:
```suggestion
                LOGGER.info("The chosen Look and Feel (that is non-standard) must be loaded now...");
```
(or: look and feel)

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
```diff
@@ -260,13 +281,25 @@ private void setLookAndFeel(String lnfName) {
...
+                // TODO - load the required LaF and defer the rest?
```
ahadas, line 291: ok

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
```diff
@@ -260,13 +281,25 @@ private void setLookAndFeel(String lnfName) {
...
+            }
 
             // Restores the contextual ClassLoader.
             currentThread.setContextClassLoader(oldLoader);
 
             mainFrames.forEach(MainFrame::updateFileTablesHeaderRenderer);
-            mainFrames.forEach(SwingUtilities::updateComponentTreeUI);
+            mainFrames.forEach(e -> {
+                SwingUtilities.updateComponentTreeUI(e.getJFrame());
+            });
```
ahadas, line 302: how about mapping each MainFrame to JFrame and then proceed with the forEach?

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeReader.java
```diff
@@ -596,203 +666,256 @@ else if(state == STATE_QUICK_LIST_ITEM_SELECTED)
...
+        }
+        return availableFonts;
+    }
 
     // - Helper methods ------------------------------------------------------
     // -----------------------------------------------------------------------
+
+    private static Set<String> getAvailableFontsFromOS() {
+        return new HashSet<>(List.of(
+                        // takes long time....
```
ahadas, line 854: yeah, it makes sense to cache but the way it's implemented now, after starting the app we would never get new fonts that are installed after the app started, no?

ahadas (commented):
> @ahadas I've just rebased the code. Any thoughts what next with this PR?

hi @pskowronek, sorry for the late reply on this - I was really busy lately..
I see that unfortunately we didn't get any response on the PR for apache-felix, right?
I'm also not sure about the SMB part
but from my point of view, this PR is not a POC at this point - we see that it brings value. the question now is what to include in 1.4 - I reviewed most of the changes (95 out of 100 files), suggested to separate some things to different PRs so that we can merge them and make this one a bit smaller
how about posting those things separately and continue to talk about the comments inside, and I'll complete the review in the next couple of days? I hope I'll be able to spend some time this week on muCommander so that we can polish this PR as well as the remaining items for 1.4 and release it at the beginning of 2024

src/main/java/com/mucommander/main/AutoProcessor.java
```diff
@@ -219,20 +255,36 @@ else if ((b != null) && actionList.contains(AUTO_DEPLOY_UPDATE_VALUE))
...
+                    try {
+                        long pre = System.currentTimeMillis();
+                        var bundle = bundleFuture.get();
+                        System.out.println("------ get bundle " + (System.currentTimeMillis() - pre) + "ms " + bundle);
+                        var startLvl = startLevel;
+                        executor.execute(() -> {
+                            try
+                            {
+                                if (bundle != null) {
+                                    if (!isFragment(bundle)) {
```
ahadas, line 274:
```suggestion
                                    if (bundle != null && !isFragment(bundle)) {
```

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas, line 1: overall, the changes here look good except for the special handling of hard-coded bundles
it would have been great to get feedback from developers of apache-felix on this but https://github.com/apache/felix-dev/pull/236 didn't get any feedback for quite some time...

mucommander-core/build.gradle
```diff
@@ -65,6 +67,7 @@ jar {
         'Import-Package': 'org.violetlib.aqua;resolution:=dynamic,com.formdev.flatlaf;resolution:=dynamic,com.apple.*;resolution:=dynamic,sun.security.action;resolution:=dynamic,*',
         'Export-Package':
             'com.mucommander.core.desktop,' +
+            'com.mucommander.preload,' +
```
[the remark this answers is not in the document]
pskowronek: for `command-core` and `mucommander-commons-runtime` (due to abusing OSVersion to initiated pre-load)

mucommander-core/src/main/java/com/mucommander/ui/dnd/ClipboardNotifier.java
```diff
@@ -49,7 +49,11 @@ public ClipboardNotifier(Action action) {
         this.action = action;
 
         // Toggle initial state
-        toggleActionState();
+        // commented out since I think it is not needed here and checking clipboard on start-up may introduce delay
```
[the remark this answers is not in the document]
pskowronek: "commented out since I think it is not needed here" - it works without that, I don't think it breaks anything.

mucommander-core/src/main/java/com/mucommander/ui/main/frame/DefaultMainFramesBuilder.java
```diff
@@ -26,6 +26,7 @@
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import org.jetbrains.annotations.NotNull;
```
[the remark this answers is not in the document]
pskowronek: oops, my mistake. For now I will remove that annotation (`javax.annotation.Nonnull` is not available just like that, don't want to add more to this PR)

mucommander-protocol-adb/src/main/java/com/mucommander/commons/file/protocol/adb/AndroidMenu.java
```diff
@@ -98,27 +99,41 @@ public void menuSelected(MenuEvent e) {
...
-        List<String> androidDevices = AdbUtils.getDevices();
-        if (androidDevices == null) {
-            setEnabled(false);
-            setToolTipText(Translator.get("adb.android_disabled"));
-            return;
-        }
-        setEnabled(true);
-        if (androidDevices.isEmpty()) {
-            add(new JMenuItem(Translator.get("adb.no_devices"))).setEnabled(false);
+        if (searching) {
```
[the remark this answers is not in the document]
pskowronek: good catch! I don't remember how it might have happened!

mucommander-core/src/main/java/com/mucommander/Application.java
```diff
@@ -163,12 +165,13 @@ private void printFileError(String msg, Exception exception, boolean quit) {
...
-        if (splashScreen != null) {
+    private void printStartupMessage(CompletableFuture<SplashScreen> splashScreenProvider,
+                                     String message) {
+        splashScreenProvider.thenAccept(splashScreen -> {
             splashScreen.setLoadingMessage(message);
-        }
+        });
 
-        LOGGER.trace(message);
+        LOGGER.error(message);
```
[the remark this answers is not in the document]
pskowronek: it was for PoC playing around

mucommander-core/src/main/java/com/mucommander/Application.java
```diff
@@ -199,8 +202,13 @@ private static void migrateCommand(String useName, String commandName, String al
     }
 
     private void run() {
-        SplashScreen splashScreen = null;
+        ExecutorService executor = Executors.newFixedThreadPool(12);
+
+        System.out.println(new Date() + " -- Application#run");
```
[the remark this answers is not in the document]
pskowronek: yup, but before merge - let's keep that and remember about it to remove just before merge

mucommander-core/src/main/java/com/mucommander/Application.java
```diff
@@ -361,141 +371,190 @@ private void run() {
...
+                printStartupMessage(splashScreenProvider, "Loading file associations..."); // TODO Localize those messages.....
+                try {
+                    com.mucommander.command.CommandManager.loadCommands();
+                } catch (Exception e) {
+                    printFileError("Could not load custom commands", e, activator.fatalWarnings());
+                }
+            });
+
+            executor.execute(() -> {
+                // Migrates the custom editor and custom viewer if necessary.
```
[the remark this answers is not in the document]
pskowronek: going to move that after CommandManager#loadCommands in 1 block

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
pskowronek: line 116: setLookAndFeel(lnfName), a couple lines above. This is case when chosen LnF is from "additional ones" - then loading is being forced (not deferred).

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
pskowronek: -> debug

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
pskowronek: in other PR

mucommander-core/src/main/java/com/mucommander/ui/main/WindowManager.java
pskowronek: later in other PR? (it's kinda overwhelming me already, especially after a couple of months from initial commit :) )

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeReader.java
pskowronek: do you mean while app is running we won't see new fonts? Or between restarts?

mucommander-core/src/main/java/com/mucommander/ui/main/FolderPanel.java
```diff
@@ -130,92 +136,101 @@ it calls actionPerformed() each time an item is highlighted with the arrow (UP/D
...
-    			new ParentFoldersQL(this),
-    			new RecentLocationsQL(this),
-    			new RecentExecutedFilesQL(this),
-    			new BookmarksQL(this),
-    			new RootFoldersQL(this),
-                new TabsQL(this)};
+        panel.setBorder(null);
+
+        JPanel locationPanel = PreloadedJFrame.getJPanel(new GridBagLayout());
+        panel.getTreeLock();
```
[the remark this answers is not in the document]
pskowronek: mistake, going to remove that

mucommander-protocol-smb/src/main/java/com/mucommander/commons/file/protocol/smb/Activator.java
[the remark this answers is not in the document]
pskowronek: The general part when smb library executes. It caught my eye in starting logs. Right now I don't remember exactly.

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: I've pigged them, and still nothing

pskowronek:
@ahadas I've just pushed fixes to address your comments. For the comments asking for "separate PR" I will do that today or tomorrow.

build.gradle
```diff
@@ -145,6 +145,8 @@ createBundlesDir.doLast {
 
 runOsgi {
     javaArgs = '-Djava.library.path=/usr/local/lib' // for Libguestfs
+    javaArgs += ' -Xshare:auto -XX:-UsePerfData -XX:+TieredCompilation -XX:TieredStopAtLevel=1' // faster start
```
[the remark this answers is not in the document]
pskowronek: #1094

mucommander-core/src/main/java/com/mucommander/core/desktop/DesktopManager.java
```diff
@@ -218,8 +218,11 @@ public static void init(boolean install) throws DesktopInitialisationException {
                 desktop = current;
                 LOGGER.debug("Using desktop: " + desktop);
                 desktop.init(install);
-                setTrashProvider(desktop.getTrash());
-                setNotifier(desktop.getNotifier());
+                new Thread(() -> {
+                    // takes a bit longer, lets do that async
+                    setTrashProvider(desktop.getTrash());
+                    setNotifier(desktop.getNotifier());
+                }, "DesktopInit").start();
```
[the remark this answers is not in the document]
pskowronek: #1095

mucommander-core/src/main/java/com/mucommander/ui/action/AbstractActionDescriptor.java
```diff
@@ -118,8 +118,6 @@ private static ImageIcon getStandardIcon(ActionId actionId) {
 
         // Look for an icon image file with the /action/<action id>.png path and use it if it exists
         iconPath = getStandardIconPath(actionId);
-        if(ResourceLoader.getResourceAsURL(iconPath) == null)
-            return null;
```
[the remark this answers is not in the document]
pskowronek: #1096

mucommander-core/src/main/java/com/mucommander/ui/icon/IconManager.java
```diff
@@ -104,7 +104,7 @@ private IconManager() {}
     public static ImageIcon getIcon(String iconPath, float scaleFactor) {
         URL resourceURL = ResourceLoader.getResourceAsURL(iconPath);
         if(resourceURL==null) {
-            LOGGER.debug("Warning: attempt to load non-existing icon: "+iconPath+" , icon missing ?");
+            LOGGER.debug("Warning: attempt to load non-existing icon: {} , icon missing ?", iconPath);
```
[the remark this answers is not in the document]
pskowronek: #1097

mucommander-core/src/main/resources/images/file/android.png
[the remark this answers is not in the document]
pskowronek: #1098

mucommander-protocol-adb/src/main/java/com/mucommander/commons/file/protocol/adb/AndroidMenu.java
```diff
@@ -98,27 +99,41 @@ public void menuSelected(MenuEvent e) {
...
-            menuItem.setText(name == null ? serial : name);
-            menuItem.setIcon(IconManager.getIcon(IconManager.FILE_ICON_SET, "android.png"));
-
-            add(menuItem);
-        }
+
+        JMenuItem searchingItem = add(new JMenuItem(Translator.get("loading")));
+        searchingItem.setEnabled(false);
+        CompletableFuture.runAsync(() -> {
+            var androidDevices = AdbUtils.getDevices();
```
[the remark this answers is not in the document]
pskowronek: @ahadas However, I don't get that. Can you pls elaborate a bit?

pskowronek:
> @ahadas I've just pushed fixes to address your comments. For the comments asking for "separate PR" I will do that today or tomorrow.

@ahadas 
Separate PR created, ping me if I missed any

pskowronek:
@ahadas rebased after merges of separated PRs

pskowronek:
@ahadas any chance to move on with the rest?

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeManager.java
```diff
@@ -57,14 +64,18 @@
...
 
     // - Class variables -----------------------------------------------------------------
     // -----------------------------------------------------------------------------------
     /** Path to the user defined theme file. */
     private static       AbstractFile userThemeFile;
     /** Default user defined theme file name. */
     private static final String       USER_THEME_FILE_NAME             = "user_theme.xml";
+
+    /** The location of available fonts list as getting them from OS takes long time (impacts startup) */
+    private static final String       AVAILABLE_FONT_CACHE             = "available_fonts_cache.cache";
```
ahadas, line 77: did you consider storing this cache in the snapshot file instead?

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeReader.java
```diff
@@ -33,84 +40,88 @@
...
-    private static final int STATE_READ_ONLY_SELECTED       = 45;
+    private static final int STATE_READ_ONLY = 43;
+    private static final int STATE_READ_ONLY_NORMAL = 44;
+    private static final int STATE_READ_ONLY_SELECTED = 45;
 
+    /** Cache of available fonts (it is very slow to initialize):
+     * https://www.mail-archive.com/java2d-interest@capra.eng.sun.com/msg02877.html,
+     * https://stackoverflow.com/questions/3237941/swing-load-available-font-family-slow-down-the-performance
+     */
+    private static Set<String> availableFonts = null;
```
ahadas, line 115:
```suggestion
    private static Set<String> availableFonts;
```

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeReader.java
ahadas: I meant between restarts
I think it's ok though if we'll store the cache in the snapshot file

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeReader.java
```diff
@@ -596,203 +666,256 @@ else if(state == STATE_QUICK_LIST_ITEM_SELECTED)
...
     // - Helper methods ------------------------------------------------------
     // -----------------------------------------------------------------------
+
+    private static Set<String> getAvailableFontsFromOS() {
+        return new HashSet<>(List.of(
+                        // takes long time....
+                        GraphicsEnvironment.getLocalGraphicsEnvironment().getAvailableFontFamilyNames())
+                .stream()
+                .map(String::toLowerCase)
+                .collect(Collectors.toList()));
```
ahadas, line 858: why not to collect it to Set?

mucommander-protocol-adb/src/main/java/com/mucommander/commons/file/protocol/adb/AndroidMenu.java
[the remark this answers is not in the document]
ahadas: I just meant to ask whether the intention of this change was to separate this part out so it won't run by the thread of Swing

src/main/java/com/mucommander/main/muCommander.java
```diff
@@ -231,15 +234,27 @@ public static void main(String[] args) throws Exception
...
 
+        new Thread(() -> {
+            System.out.println("Preloading fonts in JVM...");
+            var pre = System.currentTimeMillis();
+            try {
+                GraphicsEnvironment.getLocalGraphicsEnvironment().getAllFonts();
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+            System.out.println("Preloading fonts completed in: " + (System.currentTimeMillis() - pre) + "ms");
```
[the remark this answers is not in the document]
ahadas: partially - it address the first part of how much time it takes but it's not clear to me why not to load the fonts in a lazy way by the Theme editor

ahadas:
@pskowronek thanks again for keep rebasing and polishing this PR, it has been tough on my side to give it proper attention
I think we discussed it already but I can't recall the button line for why we trigger the initialization of the preloaded Swing components from the 'runtime' bundle rather than making 'mucommander-core-preload' a bundle and initialize it from its activator.. shall we compare the startup time in both cases?
the other comments are relatively minor

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeManager.java
pskowronek: > did you consider storing this cache in the snapshot file instead?

nope, it didn't came to my mind back then

mucommander-core/src/main/java/com/mucommander/ui/theme/ThemeReader.java
pskowronek: good catch

pskowronek:
> Swing components from the 'runtime' bundle rather than making 'mucommander-core-preload' a bundle and initialize it from its activator.. shall we compare the startup time in both cases?

I've just answered that a moment ago (somewhere above :D). OSGi framework needs time to init itself, we will loose time when we do that as bundle. In other words - we use the time OSGi fw needs to warm up to warm up swing in the meantime.

Additionally - regarding font caching and your suggestion to use snapshot - I'm still thinking if font caching is necessary since I finally added a code to load them ahead of anything in muCommander.java:24, in such case fonts will be cached by jvm already.

ahadas:
> > Swing components from the 'runtime' bundle rather than making 'mucommander-core-preload' a bundle and initialize it from its activator.. shall we compare the startup time in both cases?
> 
> I've just answered that a moment ago (somewhere above :D). OSGi framework needs time to init itself, we will loose time when we do that as bundle. In other words - we use the time OSGi fw needs to warm up to warm up swing in the meantime.

right, but even with the posted changes we'll wait for mucommander-commons-runtime bundle to load, no?

> Additionally - regarding font caching and your suggestion to use snapshot - I'm still thinking if font caching is necessary since I finally added a code to load them ahead of anything in muCommander.java:24, in such case fonts will be cached by jvm already.

you probably meant line 241 - yep, looks like we can drop the caching then

pskowronek:
> > > Swing components from the 'runtime' bundle rather than making 'mucommander-core-preload' a bundle and initialize it from its activator.. shall we compare the startup time in both cases?
> > 
> > 
> > I've just answered that a moment ago (somewhere above :D). OSGi framework needs time to init itself, we will loose time when we do that as bundle. In other words - we use the time OSGi fw needs to warm up to warm up swing in the meantime.
> 
> right, but even with the posted changes we'll wait for mucommander-commons-runtime bundle to load, no?

Sorry for confusion, I had to rewind the history of commits and conversations (I guess I mixed in my mind this and font pre-load that is being done in muCommander#main). Yes, `mucommander-commons-runtime` must be loaded then it will execute pre-load - so, if you give me a hint how to initiate pre-load in commons-runtime then I will do it.

> 
> > Additionally - regarding font caching and your suggestion to use snapshot - I'm still thinking if font caching is necessary since I finally added a code to load them ahead of anything in muCommander.java:24, in such case fonts will be cached by jvm already.
> 
> you probably meant line 241 - yep, looks like we can drop the caching then

yes. I remove the cache - do you want me to do this in this PR? If so, I will do it this evening.

ahadas:
> > > > Swing components from the 'runtime' bundle rather than making 'mucommander-core-preload' a bundle and initialize it from its activator.. shall we compare the startup time in both cases?
> > > 
> > > 
> > > I've just answered that a moment ago (somewhere above :D). OSGi framework needs time to init itself, we will loose time when we do that as bundle. In other words - we use the time OSGi fw needs to warm up to warm up swing in the meantime.
> > 
> > 
> > right, but even with the posted changes we'll wait for mucommander-commons-runtime bundle to load, no?
> 
> Sorry for confusion, I had to rewind the history of commits and conversations (I guess I mixed in my mind this and font pre-load that is being done in muCommander#main). Yes, `mucommander-commons-runtime` must be loaded then it will execute pre-load - so, if you give me a hint how to initiate pre-load in commons-runtime then I will do it.
> 
> > > Additionally - regarding font caching and your suggestion to use snapshot - I'm still thinking if font caching is necessary since I finally added a code to load them ahead of anything in muCommander.java:24, in such case fonts will be cached by jvm already.

I don't remember the entire context but when I had to make sure mucommander-core is loaded after all other bundles are loaded, I set it with level=2 [here](https://github.com/mucommander/mucommander/blob/master/src/main/java/com/mucommander/main/muCommander.java#L290), maybe we can introduce another level for the preloaded stuff to make sure it's the first bundle to be loaded

> > 
> > 
> > you probably meant line 241 - yep, looks like we can drop the caching then
> 
> yes. I remove the cache - do you want me to do this in this PR? If so, I will do it this evening.

yes please

pskowronek:
> I don't remember the entire context but when I had to make sure mucommander-core is loaded after all other bundles are loaded, I set it with level=2 [here](https://github.com/mucommander/mucommander/blob/master/src/main/java/com/mucommander/main/muCommander.java#L290), maybe we can introduce another level for the preloaded stuff to make sure it's the first bundle to be loaded

@ahadas,
1st commit:
Added Activator to `mucommander-core-preload`, set it as a dependency for `mucommander-commons-runtime`. It is being started as the first bundle.

> > yes. I remove the cache - do you want me to do this in this PR? If so, I will do it this evening.
> 
> yes please

done in 2nd commit

src/main/java/com/mucommander/main/muCommander.java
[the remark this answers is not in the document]
ahadas: I'm ok with follow-up changes, and I trust you for posting them, but it will be a challenge to track the things we say we'll follow up on.. 
but specifically about this part, does the current code also load all the fonts?

mucommander-core/src/main/java/com/mucommander/Application.java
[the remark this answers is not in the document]
pskowronek: @ahadas so, the code as it now is OK for you?

src/main/java/com/mucommander/main/muCommander.java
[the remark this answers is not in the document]
pskowronek: @ahadas Using [GraphicsEnvironment](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/awt/GraphicsEnvironment.html) you cannot load only desired fonts (correct me if I'm wrong), one need to call #getAllFonts() - and this is long process (the longer the more fonts installed in the OS). 

"and I trust you for posting them" - I'm kinda lost here, too - I don't think this requires additional changes.
A suggestion: If you spot that something needs to be changed after merge, then pls create an issue and I will address it?
I remember that I need to remove some log.error entries (or switch them to .debug) - I've marked such places with // TODO. For System.println while initializing OSGi (loading bundles), I would leave them as they are.

mucommander-core/src/main/java/com/mucommander/ui/icon/IconManager.java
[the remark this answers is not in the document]
ahadas: @pskowronek I meant:
```suggestion
            LOGGER.debug("Warning: attempt to load non-existing icon: {} , icon missing?", iconPath);
```

mucommander-core/src/main/java/com/mucommander/Application.java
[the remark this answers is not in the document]
ahadas: same here, please remove - if you still need this, please separate it out to a different commit that is not posted in order to minimize changes in this PR

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas: > I may try to drop this deferring and see how it impacts the load time.

in my tests, I didn't see an issue with deferring `gdrive` even when I had my google drive loaded on startup but @pskowronek did you check the above?

mucommander-core/src/main/java/com/mucommander/ui/dnd/ClipboardNotifier.java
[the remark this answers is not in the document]
ahadas: @pskowronek IIUC #1137 is about improving the discovering of the clipboard state, right? I don't want this to block the release of v1.4 - it's ok to keep handling the clipboard state in v1.4 as we did so far in v1.4 and handle #1137 later on

ahadas:
@pskowronek thanks again for keep pushing this one, let's polish it as I'm starting to work on creating the v1.4 release

pskowronek:
@ahadas PR updated, I hope I addressed all your comments :)

mucommander-core/src/main/java/com/mucommander/Application.java
```diff
@@ -361,141 +366,186 @@ private void run() {
...
-                    .getVariable(MuPreference.ENABLE_SYSTEM_NOTIFICATIONS,
-                            MuPreferences.DEFAULT_ENABLE_SYSTEM_NOTIFICATIONS)) {
-                printStartupMessage(splashScreen, "Enabling system notifications...");
-                if (com.mucommander.ui.notifier.NotifierProvider.isAvailable())
-                    com.mucommander.ui.notifier.NotifierProvider.getNotifier().setEnabled(true);
-            }
+            // Invoke in a different thread: https://www.oracle.com/technical-resources/articles/javase/swingworker.html
+            Thread mainThread = new Thread(() -> {
+                LOGGER.debug("muC UI about to be presented");
+                printStartupMessage(splashScreenProvider, "Loading theme...");
```
ahadas, line 490: this line should drop (we have it at line 375 now)

ahadas (commented):
@pskowronek looks great now, a minor comment inside
please update the title and description of the PR, and document the change in the readme file

pskowronek:
> @pskowronek looks great now, a minor comment inside please update the title and description of the PR, and document the change in the readme file

PR updated!

ahadas (approved):
@pskowronek thanks!

PR #1067, opened 2023-09-25
Shrink the size of jediterm fat-jar

ahadas:


pskowronek (commented):
A shy question - how come this made jar smaller?

ahadas:
@pskowronek, now that microsoft-graph also brings kotlin libraries - using them instead of bundling those libraries into the jar of jediterm reduces the size of the jar. the same goes for guava and jna, but those existed as separate bundles before adding microsoft-graph

pskowronek (approved):
Looks good

PR #1071, opened 2023-09-27
make Application#splashScreen a local variable

ahadas:


pskowronek (commented):
But if we want to have Splash Screen loaded in background (#1044) this doesn't help.
Instead of it, a completable future would be required to use Splash Screen only if already computed. I can add this to #1044. 

btw, do you want to commit this change to master?

ahadas:
@pskowronek right, it's not a solution of making things work in parallel - for this, we'll need a mechanism as you suggest. I just didn't like the part that sets the class member to null - with this change we can avoid that

yeah, let's get this one in and then modify it in #1044

PR #1090, opened 2023-12-31
Update of BinEd library

hajdam:
Update for never version of bined library and dependencies


mucommander-viewer-binary/lib/paged_data-0.2.1-SNAPSHOT.jar
ahadas, line 1: great!

ahadas (approved):
thanks @hajdam !

Issue #1062, opened 2023-09-21
[Bug] sftp: location bar does not use ssh key


weberjn:
### Is there an existing issue for this?

- [X] I have searched the existing issues (including the closed ones)

### Description

If you are connected to a server with an ssh key, and append a subdirectory in the location bar, muCommander shows a login prompt. Instead it should just change to the subdirectory.

### Steps to reproduce

append a subdirectory in the location bar and press enter

### Expected behavior

enter sub dir

### Actual behavior

show login prompt

### muCommander version

Version: 1.3.0 (Portable)
Build date: 2023/07/08
Build number: dddabaf

### Java version

Runtime version: 11.0.18

### Operating System type and version

Windows 11

ShayArtzi:
I was able to reproduce that and I'm working on a PR to address the issue, I'll post it here once it is ready for review

ShayArtzi:
PR: https://github.com/mucommander/mucommander/pull/1104

ahadas:
@ShayArtzi why do we need special handling for SFTP with a private key? I would expect that we use the credential for the remote server also in case of password authentication and also for other protocol, no?

ShayArtzi:
seems to me you are 100% right but I since I'm not super familiar with all of the different use cases I figured I'd minimize the change. I'm actually not sure if we have more protocols that support PK authentication, but LMK if you want me to take out the protocol check

ahadas:
Version 1.4.0 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.4.0-1)

PR #1104, opened 2024-02-03
Skip re-authentication for SFTP address change when using private key

ShayArtzi:
The PR addresses issue https://github.com/mucommander/mucommander/issues/1062.

When a user is connecting to an SFTP server using a private key, changing the location by typing into the address bar (vs navigating in the file panel) triggers an authentication popup.

the `BrowseLocationThread::isSftpWithPrivateKey` method detects this scenario so it can be excluded from re-authentication.


mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java
```diff
@@ -548,4 +550,15 @@ private void showDownloadDialog(AbstractFile file) {
         // Show confirmation/path modification dialog
         new DownloadDialog(mainFrame, fileSet).showDialog();
     }
+
+    private boolean isSftpWithPrivateKey() {
+        boolean sftpWithPrivateKey = false;
+        CredentialsMapping[] matchingCredentials = CredentialsManager.getMatchingCredentials(this.folderURL);
+        if (matchingCredentials.length > 0) {
```
pskowronek, line 557: iterate over the returned array?

mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java
```diff
@@ -548,4 +550,15 @@ private void showDownloadDialog(AbstractFile file) {
...
+            boolean isSftp = matchingCredentials[0].getRealm().getScheme().equals("sftp");
```
pskowronek, line 558: boolean -> var?

mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java
ShayArtzi: Good idea - thanks!

mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java
ShayArtzi: It's a matter of style, personally I think `var` promotes write-only style of coding (it's convenient when you write the code but not when you read it). However I can change if that is preferred in the context of this project

pskowronek: > It's a matter of style, personally I think `var` promotes write-only style of coding (it's convenient when you write the code but not when you read it). However I can change if that is preferred in the context of this project

I don't insist, however I think the code is more clearer and cleaner when `var` used in such cases.

ahadas (commented):
as I commented on #1062, I think this change should rather be generalized so that we can apply this logic on other file protocols that mucommander-core is not necessarily aware of (and in other file protocols the property might be different - like the credentials we store for dropbox and gdrive). however that would require a bigger change so I'm ok with doing it step by step and as a first step to implement it for SFTP but then:
1. let's use SFTP from [FileProtocols](https://github.com/mucommander/mucommander/blob/master/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/FileProtocols.java) rather then the string `"sftp"` explicitly
2. I think we should still try to authenticate here so that if the key is invalid, the authentication dialog would be presented. how about replacing the call to `isSftpWithPrivateKey` with code that runs before line 168 and does something like: if credentialsMapping is null and the protocol is SFTP and there are matching credentials with private-key, then authenticate with these credentials?

ShayArtzi:
> as I commented on #1062, I think this change should rather be generalized so that we can apply this logic on other file protocols that mucommander-core is not necessarily aware of (and in other file protocols the property might be different - like the credentials we store for dropbox and gdrive). however that would require a bigger change so I'm ok with doing it step by step and as a first step to implement it for SFTP but then:
> 
>     1. let's use SFTP from [FileProtocols](https://github.com/mucommander/mucommander/blob/master/mucommander-commons-file/src/main/java/com/mucommander/commons/file/protocol/FileProtocols.java) rather then the string `"sftp"` explicitly
> 
>     2. I think we should still try to authenticate here so that if the key is invalid, the authentication dialog would be presented. how about replacing the call to `isSftpWithPrivateKey` with code that runs before line 168 and does something like: if credentialsMapping is null and the protocol is SFTP and there are matching credentials with private-key, then authenticate with these credentials?

Thanks for the review!
For point no. 1 - done
For point no. 2 - Not sure if I fully understand your concern - let's break down the different scenarios: on line 171 we call `isSftpWithPrivateKey`, and assuming it returns `true`, we try to authenticate and there are two possibilities - we wither successfully authenticate (and then we are good to go) or the authentication fails. In this case we will go to the `catch` clause on line 399 and the authentication dialog will be shown to the user.

I definitely agree that can be generalized, but TBH I don't think I know about all the different use cases and scenarios and testing the change will be hard. On the other hand, the change I've made in this PR was easy to test and I was able to verify it works.

ahadas:
> Thanks for the review! For point no. 1 - done

:+1: 

> For point no. 2 - Not sure if I fully understand your concern - let's break down the different scenarios: on line 171 we call `isSftpWithPrivateKey`, and assuming it returns `true`, we try to authenticate ...

I see your comment: `// Special case: for SFTP connection that uses a private key, **no need to re-authenticate**`, and I can't find the authentication code that is executed in this case - can you point me to the line you refer to here?

> I definitely agree that can be generalized, but TBH I don't think I know about all the different use cases and scenarios and testing the change will be hard. On the other hand, the change I've made in this PR was easy to test and I was able to verify it works.

sure, let's get this one in and we'll think about how to extend the API of file protocols for this later

ShayArtzi:
> > Thanks for the review! For point no. 1 - done
> 
> 👍
> 
> > For point no. 2 - Not sure if I fully understand your concern - let's break down the different scenarios: on line 171 we call `isSftpWithPrivateKey`, and assuming it returns `true`, we try to authenticate ...
> 
> I see your comment: `// Special case: for SFTP connection that uses a private key, **no need to re-authenticate**`, and I can't find the authentication code that is executed in this case - can you point me to the line you refer to here?
> 
> > I definitely agree that can be generalized, but TBH I don't think I know about all the different use cases and scenarios and testing the change will be hard. On the other hand, the change I've made in this PR was easy to test and I was able to verify it works.
> 
> sure, let's get this one in and we'll think about how to extend the API of file protocols for this later

Sure - when I debugged the code to understand the flow, I observed the following flow:

1. `isSftpWithPrivateKey` (the new method I created) returns true, so the code flows to the next [if block](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L195)
2. `FileFactory::getFile` is called [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L212). This part is within the `try` block that takes care of showing the user the auth dialog if the auth fails
3. `FileFactory::createRawFile` is invoked and an authentication is attempted [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-commons-file/src/main/java/com/mucommander/commons/file/FileFactory.java#L564)

At this point even if the authentication fails for some reason then we are showing the normal auth dialog (because of the logic [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L419)). In my local testing however this is working and I can navigate inside the SFTP directory tree either by double-clicking a directory or changing the path directly

ahadas:
> Sure - when I debugged the code to understand the flow, I observed the following flow:
> 
> 1. `isSftpWithPrivateKey` (the new method I created) returns true, so the code flows to the next [if block](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L195)
> 2. `FileFactory::getFile` is called [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L212). This part is within the `try` block that takes care of showing the user the auth dialog if the auth fails
> 3. `FileFactory::createRawFile` is invoked and an authentication is attempted [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-commons-file/src/main/java/com/mucommander/commons/file/FileFactory.java#L564)
> 
> At this point even if the authentication fails for some reason then we are showing the normal auth dialog (because of the logic [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L419)). In my local testing however this is working and I can navigate inside the SFTP directory tree either by double-clicking a directory or changing the path directly

I see, so the aforementioned comment is a bit misleading since it's not that we don't need to call CredentialsManager::authenticate but we just want to skip the part that asks the user to specify credentials again since we already have the credentials set in CredentialsManager and therefore we can proceed to FileFactory::createRawFile

ok, so let me propose a different way to write this inline

ShayArtzi:
> > Sure - when I debugged the code to understand the flow, I observed the following flow:
> > 
> > 1. `isSftpWithPrivateKey` (the new method I created) returns true, so the code flows to the next [if block](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L195)
> > 2. `FileFactory::getFile` is called [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L212). This part is within the `try` block that takes care of showing the user the auth dialog if the auth fails
> > 3. `FileFactory::createRawFile` is invoked and an authentication is attempted [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-commons-file/src/main/java/com/mucommander/commons/file/FileFactory.java#L564)
> > 
> > At this point even if the authentication fails for some reason then we are showing the normal auth dialog (because of the logic [here](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L419)). In my local testing however this is working and I can navigate inside the SFTP directory tree either by double-clicking a directory or changing the path directly
> 
> I see, so the aforementioned comment is a bit misleading since it's not that we don't need to call CredentialsManager::authenticate but we just want to skip the part that asks the user to specify credentials again since we already have the credentials set in CredentialsManager and therefore we can proceed to FileFactory::createRawFile
> 
> ok, so let me propose a different way to write this inline

maybe it should read something like `Special case: for SFTP connection that uses a private key, no need to ask the user for credentials`

mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java
```diff
@@ -168,6 +168,8 @@ public void run() {
         if(credentialsMapping!=null) {
             newCredentialsMapping = credentialsMapping;
             CredentialsManager.authenticate(folderURL, newCredentialsMapping);
+        } else if (isSftpWithPrivateKey()) {
+            // Special case: for SFTP connection that uses a private key, no need to re-authenticate
```
ahadas, line 172: based on the top-level thread, how about changing the else-if below to:
```java
else if(!folderURL.containsCredentials() && !isSftpWithPrivateKey() && ...
```
and we can extend the comment there with: "There is no need to get credentials from users for locations that we already have non-password based credentials for in CredentialsManager, like SFTP with a private key" 

writing the text above made me wonder whether we should rather copy the private key from the URL that is presented in the folder panel to the new URL that is passed here - did you consider this option? (basically saying that if the scheme+host of the presented file matches those of the new URL that is specified, to copy the properties from the former to the latter)

ShayArtzi: I don't mind moving the (negated) condition into the existing `else if`. My though was to not make the existing condition more complex. I like to do it sometimes even if the block itself is empty, and I just use it for a comment explaining why it's a NO-OP. Please LMK if you'd like me to go forward with the change or just change the comment and leave the new `else if` clause as is.

For the other implementation option (`copy the private key from the URL that is presented in the folder panel to the new URL that is passed here`) - actually that was my original idea but it was not trivial to implement. It seems like the private key property was missing, and I didn't want to do extra manipulation to get it from the credential manager, since it already takes care of that internally.

In addition, since we always have a fallback of displaying the auth dialog on auth error, I figured it would be fine to assume that if this is an SFTP connection and PK was used, the credentials should be correct.

Reading the code, it seemed to me like the existing `else if` [block](https://github.com/mucommander/mucommander/blob/5512bbfafff784062784301330b9da15a0c3c743/mucommander-core/src/main/java/com/mucommander/core/BrowseLocationThread.java#L177) was added at some later stage as an optimization and figured I would just skip it.

ahadas: > I don't mind moving the (negated) condition into the existing `else if`. My though was to not make the existing condition more complex. I like to do it sometimes even if the block itself is empty, and I just use it for a comment explaining why it's a NO-OP. Please LMK if you'd like me to go forward with the change or just change the comment and leave the new `else if` clause as is.

ok, yeah - it would complicate the existing condition but overall it would better reflect our intention so please change the existing condition
 
> For the other implementation option (`copy the private key from the URL that is presented in the folder panel to the new URL that is passed here`) - actually that was my original idea but it was not trivial to implement. It seems like the private key property was missing, and I didn't want to do extra manipulation to get it from the credential manager, since it already takes care of that internally.

ack, we can reconsider this when we generalize this logic for other credential properties, that's fine

ShayArtzi: Sounds good - I made the change and tested locally again. We should be good to go!

ahadas (approved):
thanks @ShayArtzi!

Issue #1165, opened 2024-05-04
[Enhancement] Click "Name" to always sort files from A to Z


Andreas0602:
### Description

When clicking the "Name" column header, files should be sorted in alphabetical ascending order. Currently, this does not always occur. Here's a scenario in which muCommander does not behave as desired:

1. Make sure that the files are sorted by date in descending order (topmost file is newest).
2. Click the "Name" column header.

**Desired result:** Files are sorted in alphabetical ascending order (topmost file is aaaaaa.txt).
**Actual result:** Files are sorted in alphabetical descending order (topmost file is zzzzzz.txt). Only after clicking "Name" a second time, the sort order will be as desired.

It is not good that the sort order after clicking "Name" depends on the previous sort order of "Date".

ShayArtzi:
I can see the issue. Basically, once a sort order was established, we retain it until a reverse operation happens. Let me create a PR to address this

PR #1166, opened 2024-05-04
Reset sort order when sort column changes

ShayArtzi:
This PR addresses issue #1165.

Original state: In the file table, when clicking `column A` after the table was already sorted by `column B`, the sort order is retained. This leads to an undesired result if `column B` is sorted in a descending order (since after clicking `column A`, its sort direction will also be descending and not ascending).

State after this change: If the current sort column is different then the previously selected sort column, reset the sort order to ascending.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTable.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTable.java
@@ -857,14 +857,11 @@ public class FileTable extends JTable implements MouseListener, MouseMotionListe
     public void sortBy(Column criterion) {
         if (criterion==sortInfo.getCriterion()) {
             reverseSortOrder();
-            return;
         } else {
             // The selected column is different from the previously selected one.
             // In this case, default to ascending sort order
-            sortInfo.setAscendingOrder(true);
+            sortBy(criterion, true);
         }
-
-        sortBy(criterion, sortInfo.getAscendingOrder());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTable.java
```diff
@@ -858,6 +858,10 @@ public void sortBy(Column criterion) {
         if (criterion==sortInfo.getCriterion()) {
             reverseSortOrder();
             return;
+        } else {
+            // The selected column is different from the previously selected one.
+            // In this case, default to ascending sort order
+            sortInfo.setAscendingOrder(true);
```
ahadas, line 864: how about calling `sortBy` here with ascending==true instead? (and remove the `return` statement above which becomes redundant)

ahadas (commented):
@ShayArtzi thanks for posting this - the change looks correct, I commented about code style inside and please add a line to describe this change in the readme file

mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTable.java
ShayArtzi: In that case I will have to duplicate the next line (call to `sortBy` at line 867) - was that your intention?

ahadas: not to duplicate but to replace the call to sortBy at line 867

ShayArtzi: ohhh right - got you. let me make the change

ShayArtzi: I made the change and updated the readme file as well. Thanks for your feedback!

PR #1176, opened 2024-05-07
Open with App - faster loading (and a fix for slowish macbooks)

pskowronek:
Open with App - faster loading by first trying `bash|zsh -c "which duty"` and then `bash|zsh -l -c "which duty"`.

"-l" option is generally better, but is 10x slower than without it. I will first try without "-l" and then as a fallback with it.

Additionally, I've increased timeout for execution from 500ms to 1s. Surprisingly, on pretty new macbooks with A/V etc it may take more than 500s to execute the command when '-l' is present.

### DIFF
--- a/mucommander-os-macos/src/main/java/com/mucommander/desktop/macos/OSXDesktopAdapter.java
+++ b/mucommander-os-macos/src/main/java/com/mucommander/desktop/macos/OSXDesktopAdapter.java
@@ -479,9 +479,9 @@ public class OSXDesktopAdapter extends DefaultDesktopAdapter {
             BufferedReader stdInput = new BufferedReader(new InputStreamReader(
                     useStdErr ? proc.getErrorStream() : proc.getInputStream()));
             int exitCode = Integer.MIN_VALUE;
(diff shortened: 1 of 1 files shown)


ahadas (approved):

ahadas:
@pskowronek I forgot to ask to document this change in the readme file, could you please do that?

pskowronek:
> @pskowronek I forgot to ask to document this change in the readme file, could you please do that?

@ahadas hm, it was documented sometime ago for 1.3.0 I think: https://github.com/mucommander/mucommander/commit/5e3d56f9218ff518b786698c557774b1ffac782a

pskowronek:
> > @pskowronek I forgot to ask to document this change in the readme file, could you please do that?
> 
> @ahadas hm, it was documented sometime ago for 1.3.0 I think: [5e3d56f](https://github.com/mucommander/mucommander/commit/5e3d56f9218ff518b786698c557774b1ffac782a)

Ah, you mean to update readme to state that in bug/improvement section.

Issue #1182, opened 2024-05-10
[Bug] "muCommander" freezes at the initialization stage at the first start.


demetriuzz:
### Is there an existing issue for this?

- [X] I have searched the existing issues (including the closed ones)

### Description

Initialization window _freezes_ when `muCommander` start **first** time: when user enter name+password and run new session in `XFCE`.

### Steps to reproduce

It's not _easy_. We need to start a new session.

### Expected behavior

Need _expected_ process - run and work program.

### Actual behavior

First you need to _kill_ the `X window`: instance of `muCommander`.
Then again run `muCommander`.
But _sometimes_ you need to kill the startup window **twice**.

### Screenshots?

![изображение](https://github.com/mucommander/mucommander/assets/61899867/795b0b40-be79-49c1-ae11-3184e9c56202)
And no more ...

### muCommander version

version `1.4.0` from installed `mucommander_1.4.0_amd64.deb`.
(Build date: 2024/05/02
 Build number: a470208)

### Java version

`Java` version embedded in `muCommander`
```text
Runtime version: 23-internal
VM name: OpenJDK 64-Bit Server VM
VM version: 23-internal-ahadas-737c9028db3d95512c25dc620d5a11dcec087f91
VM vendor: Oracle Corporation
```
`Java` version on host:
```text
openjdk version "17.0.11" 2024-04-16
OpenJDK Runtime Environment (build 17.0.11+9-Debian-1deb12u1)
OpenJDK 64-Bit Server VM (build 17.0.11+9-Debian-1deb12u1, mixed mode, sharing)
```

### Operating System type and version

`XFCE 4.18`
on Linux Debian `12.5`
`Linux 6.1.0-20-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.85-1 (2024-04-11) x86_64 GNU/Linux`

### Relevant log output

```shell
Log info from console (form file `.xsession-errors` in home directory):

[muEpoch+2217ms] muCommander: OSGi framework with muC fully started
Exception in thread "MainFrameInit" java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 4
        at com.mucommander.ui.main.MainFrame.<init>(MainFrame.java:317)
        at com.mucommander.ui.main.frame.DefaultMainFramesBuilder.createMainFrame(DefaultMainFramesBuilder.java:121)
[...]

PR #1184, opened 2024-05-10
Attempt to fix launch fail

pskowronek:
Attempt to fix launch fail #1182

It might be related to recent update where things are run in parallel for faster startup, or it may not, or it may have start to manifest itself under such condition.

I was unable to reproduce, even having multiple tabs (this is only when this may occur), however I've spotted that JTabbedPane#getSelectedIndex may return -1 if no tab is selected. This index is being used in the HideableTabbedPane#updateTab (via update event).

@ahadas - I think you know the mechanics of those tabs better, can you take a look?

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/tabs/HideableTabbedPane.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/tabs/HideableTabbedPane.java
@@ -188,7 +188,10 @@ public class HideableTabbedPane<T extends Tab> extends JComponent implements Tab
 	 * @param updater - object that will be used to update the tab
 	 */
 	protected void updateCurrentTab(TabUpdater<T> updater) {
-		tabsCollection.updateTab(getSelectedIndex(), updater);
+		var idx = getSelectedIndex();
+		if (idx >= 0) {
+			tabsCollection.updateTab(idx, updater);
+		}
 	}
 
 	/* Actions that depended on the display type (single/multiple tabs) */
--- a/package/readme.txt
+++ b/package/readme.txt
@@ -42,6 +42,8 @@ Localization:
 
 Bug fixes:
 -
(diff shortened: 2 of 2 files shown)


pskowronek:
@ahadas - let's merge this fix and release minor update to muC?

mucommander-core/src/main/java/com/mucommander/ui/tabs/HideableTabbedPane.java
```diff
@@ -188,7 +188,10 @@ protected void addAndSelectTab(T tab) {
 	 * @param updater - object that will be used to update the tab
 	 */
 	protected void updateCurrentTab(TabUpdater<T> updater) {
-		tabsCollection.updateTab(getSelectedIndex(), updater);
+		var idx = getSelectedIndex();
+		if (idx >= 0) {
+			tabsCollection.updateTab(idx, updater);
+		}
```
ahadas, line 194: I've managed to reproduce it by:
1. stopping both threads that initialize the folder panels in the constructor of FileTableTabs
2. let the thread of the right folder panel continue and stop it in LocationChanger#tryChangeCurrentFolderInternal, after `mainFrame.setNoEventsMode(true);` is called
3. let the thread of the left folder panel continue

the thread of the left folder panel would raise the same exception
the explanation is that in no-event-mode, the selected index of FileTableTabbedPane is not set (see FileTableTabbedPane#setSelectedIndex) and therefore we get -1 from getSelectedIndex here

this change would solve that issue but the selected index won't be set properly
I'd prefer to see how we can avoid this situation

ahadas: posted an alternative solution in #1201

pskowronek:
I'm closing this PR (using alternative fix by @ahadas )

Issue #1187, opened 2024-05-11
[Bug] Undesired beep and undesired character when leaving terminal by pressing F12


Andreas0602:
### Is there an existing issue for this?

- [X] I have searched the existing issues (including the closed ones)

### Description

There's an undesired beep and an undesired `~` character when leaving the terminal by pressing F12.

### Steps to reproduce

1. Start muCommander
2. Press F12
3. Press F12 again

### Expected behavior

After step 3, there's no beep and no undesired character in the terminal.

### Actual behavior

After step 3, there's an undersired beep and there is the undesired character `~` in the terminal.

### muCommander version

Version: 1.5.0
Build date: 2024/05/08
Build number: 54524da

### Java version

Runtime version: 23-internal
VM name: OpenJDK 64-Bit Server VM
VM version: 23-internal-ahadas-0081ac197a656cdc0d88d5a5e2a0f399ee023654
VM vendor: Oracle Corporation

### Operating System type and version

Name: Mac OS X
Version: 14.4.1
Architecture: aarch64

### Relevant log output

```shell
none
```

pskowronek:
> When I turn it on, the problem does not occur.

on or off? And what KB settings you have, DE?

pskowronek:
remember to restart muC after changing the settings :) and pls provide final observations and steps to reproduce (like DE keyboard etc). I will try to provide the fix for your scenario.

pskowronek:
@Andreas0602 you were right, it is kinda randomish - happened to me on both settings.
Can you check this PR #1188 ? With this fix it doesn't happen to me anymore.

PR #1188, opened 2024-05-12
Terminal - fix for undesired beep and undesired character when leaving terminal

pskowronek:
Terminal - fix for undesired beep and undesired character when leaving terminal #1187

There are 2 (or even 3 problems):
1) ToggleTerminalAction is constructed a couple of times, that could lead to creation of a couple of Terminal instances - this has been addressed by adding a static handler for keeping the instance plus a static LOCK. This ensures only one Terminal is out there. The case of multiple instances could be observed when using F12 to open terminal and leaving, then clicking on vertical bar to open the terminal - an old terminal was replaced with a new copy, which in turn had no problem with undesired beep and undesired character (!)

2) The root cause why multiple instantiations of ToggleTerminalAction happens has to be solved separately (partially due to bubbling up notifications while Main Menu and Toolbar items creation - it can be related to running Toolbar and MainMenu execution in parallel?) - it seems the multiple instantiations happen magically, see comment inside PR: https://github.com/mucommander/mucommander/pull/1190

3) Observation from 1) led me to check TerminalIntegration - some gut feeling told me that there might be some race condition within Terminal itself, so I decided to execute TerminalPanel.addCustomKeyListener  first and TerminalPanel.addFocusListener a bit later. This helped to the point I cannot reproduce it on my mac and pref & KB settings.

Pending verification from Andreas.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
@@ -36,7 +36,8 @@ import com.mucommander.ui.terminal.TerminalIntegration;
  */
 public class ToggleTerminalAction extends ActiveTabAction {
 
-    private volatile TerminalIntegration terminalIntegration = null;
+    private volatile static TerminalIntegration terminalIntegration = null;
+    private final static Object LOCK = new Object();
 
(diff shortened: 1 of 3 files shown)


pskowronek:
Andreas confirmed that it fixed the problem for him

ahadas (approved):

PR #1190, opened 2024-05-13
Fixes for MuAction cache related to concurrency

pskowronek:
Fixes for MuAction cache related to concurrency in `ActionManager`.

There is a question to @ahadas inside the commit - I will remove that TODO comment in the final PR after review.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/ActionManager.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/ActionManager.java
@@ -19,17 +19,18 @@ package com.mucommander.ui.action;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 import java.util.Vector;
 import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.swing.ImageIcon;
 import javax.swing.KeyStroke;
 
+import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -61,7 +62,7 @@ public class ActionManager {
     private static final Logger LOGGER = LoggerFactory.getLogger(ActionManager.class);
 
     /** MuAction id -> factory map */
-    private static Map<ActionId, ActionFactory> actionFactories = new Hashtable<>();
+    private static Map<ActionId, ActionFactory> actionFactories = new ConcurrentHashMap<>();
 
     /** MainFrame -> MuAction map */
     private static WeakHashMap<MainFrame, Map<ActionParameters, ActionAndIdPair>> mainFrameActionsMap = new WeakHashMap<>();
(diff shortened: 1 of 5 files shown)


mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
```diff
@@ -86,6 +86,13 @@ private void ensureIntegrationIsInitialized() {
         if (terminalIntegration == null && verticalSplit != null) {
             // doing it lazy, because in c-tor main frame might not be fully built (no vertical split pane yet)
             synchronized (LOCK) {
+                // TODO I'm lost - I see in debugger that finally this class is instantiated only once,
+                // but if I put breakpoint below I can see that mysteriously terminalIntegration
+                // is being set 2 times - first time, OK, good, but second time terminalIntegration
+                // (called via fireActivePanelChanged) is null again on the same instance of ToggleTerminalAction (!) - as if
+                // terminalIntegration was reset to null by reflection or clone, or sth?
+                // In Idea's debugger on the second occurrence there's a note that ToggleTerminalAction
+                // instance "is being initialized". ToggleTerminalAction has the same id!
```
ahadas, line 95: I see a similar behavior when debugging this from Eclipse but I don't see the constructor of TerminalIntegration being called twice - it looks like when placing a breakpoint in `terminalIntegration = new TerminalIntegration(mainFrame, verticalSplit);` then for some reason we get back to it when returning from the call to the aforementioned constructor.. @pskowronek when you put a breakpoint in the constructor of TerminalIntegration, do you see it called twice?

pskowronek: > but I don't see the constructor of TerminalIntegration being called twice

Yup, that's why I was in desperation thinking that some cloned version appears or sth.

> when you put a breakpoint in the constructor of TerminalIntegration, do you see it called twice?

Yes, and that's why there some problems with Terminal (that there were 2 terminals initialized). That's why this helped in #1188: https://github.com/mucommander/mucommander/pull/1188/commits/7c94f2692abb542c10e6caed5fecdef51dd448ae#diff-0b49347bdb327817a5590941c5cdc9790e064980e06ee7801144a18a6bb644baR40

ShayArtzi: I think the `volatile` variable might not behave as expected when it is called from the constructor, like in this case (since the object is not fully initialized yet). One way to work around this might be to have some static reference to a `TerminalIntegration` instance saved somewhere else

pskowronek: > I think the `volatile` variable might not behave as expected when it is called from the constructor, like in this case (since the object is not fully initialized yet). One way to work around this might be to have some static reference to a `TerminalIntegration` instance saved somewhere else

nope, I changed:
```private volatile static TerminalIntegration terminalIntegration = null;```
to
```private TerminalIntegration terminalIntegration = null;```
and I see 2x invocations of TerminalIntegration.

I'm going to edit the comment just for state the behavior if volatile static is not used (`static` ensures that 
TerminalIntegration is not called twice).

**Comment by pskowronek** [comment]
@ahadas - let's merge this fix and release minor update to muC?

File: mucommander-core/src/main/java/com/mucommander/ui/action/ActionManager.java [hunk 2]
```diff
@@ -367,86 +368,94 @@ public static MuAction getActionInstance(ActionType actionId, MainFrame mainFram
...
-
-            // Looks for the action's factory
-            ActionFactory actionFactory = actionFactories.get(actionId);
-            if (actionFactory == null) {
-                LOGGER.error("couldn't initiate action: " + actionId + ", its factory wasn't found");
-                return null;
-            }
+        // or create one if necessary
+        var result = mainFrameActions.computeIfAbsent(
+                actionParameters, (actParam) -> createActionAndIdPair(actParam, mainFrame));
```
ahadas, line 383:
```suggestion
                actionParameters, params -> createActionAndIdPair(params, mainFrame));
```

package/readme.txt
```diff
@@ -39,6 +39,7 @@ Improvements:
 - Faster way of finding applications for Open With context menu (macOS)
 - Debug Console is able to show logs with stacktraces (if present)
 - Improved Debug Console to display logs using tree view with folded stacktraces and it can be opened in background (non-modal) with logs autorefresh
+- Better concurrency handling while initializing Actions
```
ahadas, line 42: @pskowronek  what made you do these changes? it feels like there was an issue that is solved by these, no?

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
```diff
@@ -36,6 +36,13 @@
...
 public class ToggleTerminalAction extends ActiveTabAction {
 
+    /**
+     * A reference to Terminal Integration. Made static (and volatile) to ensure
+     * that TerminalIntegration is initialized only once, otherwise, for some unknown
+     * to me reason even-though ToggleTerminalAction constructor is called once, the
+     * TerminalIntegration is called twice as if there was another instance of
+     * ToggleTerminalAction with terminalIntegration = null - but there's only one instance (!).
+     */
     private volatile static TerminalIntegration terminalIntegration = null;
```
ahadas, line 46:
```suggestion
    private volatile static TerminalIntegration terminalIntegration;
```
while on it.. and what does `the TerminalIntegration is called twice` mean, that the constructor of `TerminalIntegration` is called twice?

ahadas: > and what does `the TerminalIntegration is called twice` mean, that the constructor of `TerminalIntegration` is called twice?

ah ok, [you replied on that already](https://github.com/mucommander/mucommander/pull/1190#discussion_r1600694968)
interesting..

package/readme.txt
pskowronek: > @pskowronek what made you do these changes? it feels like there was an issue that is solved by these, no?

I think it started with Terminal beep investigation where I spotted that there were two instances of Terminal.

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
pskowronek: yup, the constructor of TerminalIntegration. Going to add 'constructor' to that comment.

pskowronek:
PR updated

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
ahadas: ok, so I took a further look at it and I see that the constructor of TerminalIntegration is called from three places during startup.

the thread that initializes the toolbar:
image

the thread that initializes the menubar:
image

and the thread that initializes the mainframe (MainFrameInit):
image

setting `terminalIntegration` as static prevents its instantiation several times but it seems problematic to have it static and initializing it with a main frame, what happens with several main frames?.. I think it would be better to "inject" the TerminalIntegration from the MainFrame

ahadas: @pskowronek could you please review the changes in the last two commits?

pskowronek: nice catch! So it was rather a problem with catching breakpoints in multithreading?

> could you please review the changes in the last two commits?
checking...

pskowronek (commented):
looks good

pskowronek (commented):
second commit looks good, too

mucommander-core/src/main/java/com/mucommander/ui/action/impl/ToggleTerminalAction.java
ahadas: > nice catch! So it was rather a problem with catching breakpoints in multithreading?

yeah, probably

pskowronek: on Eclipse I haven't had such problems, but maybe under Idea there are some caveats I don't get.
It was probably a problem with half-baked class initialized in the same time, nevertheless - interesting.

ahadas:
@pskowronek I see you've updated the branch and dropped the last 2 commits in the process, what's the reason for the update?

pskowronek:
> @pskowronek I see you've updated the branch and dropped the last 2 commits in the process, what's the reason for the update?

Forgot to pull the changes (just wanted to rebase with the latest master) :/

ahadas:
> > @pskowronek I see you've updated the branch and dropped the last 2 commits in the process, what's the reason for the update?
> 
> Forgot to pull the changes (just wanted to rebase with the latest master) :/
> 
> Do you have your commits on local machine? Sorry for trouble.

ack, no problem :) I have them locally

ahadas (approved):

pskowronek:
> > > @pskowronek I see you've updated the branch and dropped the last 2 commits in the process, what's the reason for the update?
> > 
> > 
> > Forgot to pull the changes (just wanted to rebase with the latest master) :/
> > Do you have your commits on local machine? Sorry for trouble.
> 
> ack, no problem :) I have them locally

@ahadas Thanks, are you using CLI git, or GH? Especially as it comes to pulling changes from remote branch to local branch (just like in this case).

ahadas:
> > > > @pskowronek I see you've updated the branch and dropped the last 2 commits in the process, what's the reason for the update?
> > > 
> > > 
> > > Forgot to pull the changes (just wanted to rebase with the latest master) :/
> > > Do you have your commits on local machine? Sorry for trouble.
> > 
> > 
> > ack, no problem :) I have them locally
> 
> @ahadas Thanks, are you using CLI git, or GH? Especially as it comes to pulling changes from remote branch to local branch (just like in this case).

I always use git from the CLI, got used to it..

ahadas:
Version 1.4.1 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.4.1-1)

PR #1193, opened 2024-05-15
Fix Shell panel name in preferences

pskowronek:
Fix Shell panel name in preferences:

obraz


mucommander-translator/src/main/resources/dictionary.properties
```diff
@@ -861,7 +861,7 @@ prefs_dialog.mail_address = Your email address
 prefs_dialog.mail_server = SMTP server
 prefs_dialog.misc_tab = Misc
 prefs_dialog.confirm_on_quit = Show confirmation dialog on quit
-prefs_dialog.shell = $[RunCommand.label]
+prefs_dialog.shell = Shell
```
ahadas, line 864: how about calling it "Terminal" then to reduce the amount of translation items?

pskowronek: Afair, this is for choosing bash vs zsh, so shell is appropriate I think

pskowronek: > how about calling it "Terminal" then to reduce the amount of translation items?

ok, now I'm looking normally thru laptop - yes, you are right. I will change this to Terminal

pskowronek: @ahadas  PR updated

ahadas (approved):

PR #1201, opened 2024-05-28
Prevent IndexOutOfBoundsException on startup

ahadas:
Now that we initialize both folder panels in parallel, it could be that one thread sets no-events-mode on the MainFrame and as a result, the selected index is not set for the FileTableTabbedPane of the other FolderPanel (see FileTableTabbedPane#setSelectedIndex).

A solution for this is not to set the no-events-mode while the MainFrame is not displayed - this way the selected index for each FileTableTabbedPane is set on startup, before the MainFrame is displayed, and yet we block other operations from starting at the same time later on when the MainFrame is displayed as we did before.


pskowronek (approved):
looks good

PR #1202, opened 2024-05-29
Clean up things I've noticed during recent reviews

ahadas:


mucommander-core/src/main/java/com/mucommander/ui/dialog/debug/DebugConsoleDialog.java
```diff
@@ -328,29 +325,22 @@ private Color getColorForLevel(LogLevel level) {
...
+                    return Color.BLACK;
                 case FINE:
-                    color = Color.DARK_GRAY;
-                    break;
+                    return Color.DARK_GRAY;
                 default:
                     // Between Color.GRAY and Color.DARK_GRAY
-                    color = new Color(110, 110, 110);
-            };
-            return color;
```
pskowronek, line 353: matter of taste, personally I like single point of return (unless it is return on the beginning of method - args check etc). Just a comment.

pskowronek (approved):

mucommander-core/src/main/java/com/mucommander/ui/dialog/debug/DebugConsoleDialog.java
ahadas: yeah, that's subjective. I tend to prefer returning from each case when the switch-case is the very last element in a method

Issue #1203, opened 2024-05-31
File dates inside zip file all show as 1/1/80


rhstanton:
### Is there an existing issue for this?

- [X] I have searched the existing issues (including the closed ones)

### Description

When I look at the contents of a zip file recently downloaded from Google Drive, muCommander (wrongly) shows all file dates as 1/1/80, and this date is transferred across if I copy the files inside muCommander.

### Steps to reproduce

I just downloaded some files from Google Drive. Google zipped them and saved them in a zip file on my local disk. 

When I click on the zip file inside muCommander to view the contents, all the file dates inside the zip file show up as 1/1/80, and if I copy them to another directory, the resulting files are also dated 1/1/80. In reality, all the files are dated 5/28/24, and this date shows up if I list the archive contents using ```unzip -l```. 

I also tested the archive using ```unzip -t``` and it reports no errors.

This didn't happen with a  simple manually created zip file just now, so let me know if you would like a copy of the zip file in question.

### Expected behavior

The file dates should display and copy correctly.

### Actual behavior

File dates display and are copied as 1/1/80.

### muCommander version

1.4.0

### Java version

Runtime version: 23-internal
VM name: OpenJDK 64-Bit Server VM
VM version: 23-internal-ahadas-0081ac197a656cdc0d88d5a5e2a0f399ee023654
VM vendor: Oracle Corporation

### Operating System type and version

Name: Mac OS X
Version: 12.7.5
Architecture: x86_64

ShayArtzi:
Hi @rhstanton, it would be helpful if you could attach an example file. Thanks!

rhstanton:
Sure. I just uploaded two small text files to Google Drive then downloaded them both, creating a zip file with the exact same issue, which I've attached.

[Test-20240531T050145Z-001.zip](https://github.com/user-attachments/files/15508509/Test-20240531T050145Z-001.zip)

ShayArtzi:
Thanks - that helps. I was able to reproduce the issue on my machine and diagnose the cause. When muC parses a zip file to list its content, is refers to the classic modification time and date fields for each file. However, in some cases those (now legacy) fields are not being used. Instead, the header includes an `extra field` (`0x5455`) which stores an extended timestamp.

In this case, the legacy date/time fields defaults to 1980-01-01 (this is MS-DOS formats which stores date as an offset from 1980).

[...]

PR #1204, opened 2024-06-01
Extended timestamp support for zip files

ShayArtzi:
This PR address issue #1203.

## Background
A zip archive contains a central directory record, which lists all files included in the archive file. For each file there is a modification time and date fields (in MS-DOS format, meaning offset in seconds from 1980-01-01).

In some modern zip implementation (for example, Google Drive generated files), instead of using the above fields, an extended timestamp field is being used. This field is included in the _Extra field_ of a file record (see [reference](https://users.cs.jmu.edu/buchhofp/forensics/formats/pkzip.html) and also on [libzip's website](https://libzip.org/specifications/extrafld.txt)). This field uses epoch time. In these cases, muC will show the file modification time as 1980-01-01, since the MS-DOS time offset is 0.

## Suggested change
muC uses custom code to parse zip files (based on Apache Ant). After investigating the code, I saw _Extra field_ values are already parsed in a generic manner (represented as `UnrecognizedExtraField`). I have created a new extra field implementation called `ExtendedTimestampExtraField` to represent the aforementioned extended timestamp field.

I then altered the `ZipEntry::getTime` method to use the extended time data if it is available. 

## Testing
Using a zip file generated by Google Drive and provided by @rhstanton [here](https://github.com/mucommander/mucommander/issues/1203#issuecomment-2141238123), here is a screenshot before the change (notice file dates):
image

and after the change:
image

### DIFF
--- a/mucommander-format-zip/src/main/java/com/mucommander/commons/file/archive/zip/provider/ExtendedTimestampExtraField.java
+++ b/mucommander-format-zip/src/main/java/com/mucommander/commons/file/archive/zip/provider/ExtendedTimestampExtraField.java
@@ -1,0 +1,63 @@
+package com.mucommander.commons.file.archive.zip.provider;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+
+public class ExtendedTimestampExtraField implements ZipExtraField {
+
+    private static final ZipShort ID = new ZipShort(0x5455);
+    private static final byte FLAG_MODTIME_PRESENT = 0x01;
+
+    private byte[] localData;
+
+    private Long javaTime;
+
+    public Long getJavaTime() {
+        return javaTime;
+    }
+
+    @Override
+    public ZipShort getHeaderId() {
+        return ID;
+    }
+
+    @Override
+    public ZipShort getLocalFileDataLength() {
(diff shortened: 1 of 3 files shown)


ahadas (commented):
nice! minor comments inside

ShayArtzi:
> nice! minor comments inside

thanks for the review! After making the changes I tested the code again to make sure it works with both types of zip files (using legacy and extended timestamp fields)

ahadas (approved):
thanks @ShayArtzi! I pushed another commit of minor cleanup

ShayArtzi:
I forgot to handle the case where `getTimeExtended` can return null - it should return an empty optional

ahadas:
> I forgot to handle the case where `getTimeExtended` can return null - it should return an empty optional

oh right, could you post a fix please?

ShayArtzi:
> > I forgot to handle the case where `getTimeExtended` can return null - it should return an empty optional
> 
> oh right, could you post a fix please?

just did ;)

ahadas:
> > > I forgot to handle the case where `getTimeExtended` can return null - it should return an empty optional
> > 
> > 
> > oh right, could you post a fix please?
> 
> just did ;)

you had a typo in the commit message so I fixed it and changed getExtendedTime a bit

PR #1206, opened 2024-06-05
Reuse translations of 'Terminal'

ahadas:
introduce an entry called 'terminal' in the dictionary that stores the translation for 'Terminal' and other places that refer to the terminal point to, and convert the entry theme_editor.terminal_tab in dictionary files in which it holds the translation of 'Terminal' to this new 'terminal' entry.


pskowronek (approved):
looks good

PR #1213, opened 2024-06-13
Update Gradle, TestNG, JUnit, junrar, commons-lang3, commons-collecti…

olivierpaul:
…ons4, commons-logging


pskowronek (approved):
looks good from side (and builds & runs on macOS)

ahadas (approved):

PR #1220, opened 2024-06-28
image viewer - faster display of big images (esp JPGs), some smaller fixes, additional formats

pskowronek:
- faster load of bigger JPGs (like 10MB+ - ~2-3x faster) - thanks to TwelveMonkeys and its `ByteArrayImageInputStream`, plus its JPEG/TIFF plugins 
- fix for JPEG orientation, apparently TwelveMonkeys JPEG plugin works better to determine the correct orientation
- fix for non-deterministic behavior when loading image and window was empty
- added TwelveMonkeys plugins for additional image formats that image viewer can display: webp, psd (and psb)
- the list of extensions supported by JRE TwelveMonkeys plugins is taken from ImageIO and not hardcoded

TwelveMonkeys could also be used for nice resampling for scaled images (TBD in future).

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/layout/AsyncPanel.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/layout/AsyncPanel.java
@@ -26,6 +26,7 @@ import java.awt.Window;
 import javax.swing.JComponent;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
+import javax.swing.SwingUtilities;
 import javax.swing.border.EmptyBorder;
 import javax.swing.event.AncestorEvent;
 import javax.swing.event.AncestorListener;
@@ -103,19 +104,15 @@ public abstract class AsyncPanel extends JPanel {
      * Loads the target component by calling {@link #getTargetComponent()} and replace the wait component by it.
      */
     private void loadTargetComponent() {
-        new Thread() {
-            @Override
-            public void run() {
-                JComponent targetComponent = getTargetComponent();
-
+        new Thread(() -> {
+            JComponent targetComponent = getTargetComponent();
+            SwingUtilities.invokeLater(() -> {
                 remove(waitComponent);
                 setBorder(new EmptyBorder(0, 0, 0, 0));
-
                 add(targetComponent, BorderLayout.CENTER);
-
                 updateLayout();
-            }
-        }.start();
+            });
+        }).start();
     }
 
     /**
--- a/mucommander-viewer-image/build.gradle
+++ b/mucommander-viewer-image/build.gradle
@@ -6,17 +6,35 @@ dependencies {
     api project(':mucommander-viewer-api')
     api project(':mucommander-translator')
 
+    comprise 'com.twelvemonkeys.imageio:imageio-core:3.11.0'
+
+    // image plugins to handle additional types (like psd) or faster alternatives to built-in ones (jpeg)
+    comprise 'com.twelvemonkeys.imageio:imageio-jpeg:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-tiff:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-psd:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-webp:3.11.0'
+    // requirements for plugins
+    comprise 'com.twelvemonkeys.common:common-lang:3.11.0'
+    comprise 'com.twelvemonkeys.common:common-io:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-metadata:3.11.0'
+
+
     compileOnly 'com.google.code.findbugs:jsr305:1.3.9'
 
     testImplementation 'org.testng:testng:7.10.2'
 }
 
 jar {
+   from configurations.comprise.collect { it.isDirectory() ? it : zipTree(it).matching {
+        exclude 'META-INF/**'
+   } }
+   duplicatesStrategy = DuplicatesStrategy.EXCLUDE
    bnd ('Bundle-Name': 'muCommander-viewer-image',
         'Bundle-Vendor': 'muCommander',
-        'Bundle-Description': 'Library for hexadecimal viewer/editor',
+        'Bundle-Description': 'Library for previewing image files',
         'Bundle-DocURL': 'https://www.mucommander.com',
         'Export-Package': 'com.mucommander.viewer.image',
+        'Import-Package': '!magick,*',
         'Bundle-Activator': 'com.mucommander.viewer.image.Activator',
         'Specification-Title': "muCommander",
         'Specification-Vendor': "Arik Hadas",
--- a/mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/Activator.java
+++ b/mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/Activator.java
@@ -18,10 +18,19 @@ package com.mucommander.viewer.image;
 
 import com.mucommander.snapshot.MuSnapshot;
 import com.mucommander.viewer.FileViewerService;
-import javax.annotation.ParametersAreNonnullByDefault;
+import com.twelvemonkeys.imageio.plugins.jpeg.JPEGImageReaderSpi;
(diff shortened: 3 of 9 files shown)


mucommander-core/src/main/java/com/mucommander/ui/layout/AsyncPanel.java
```diff
@@ -103,19 +104,15 @@ public void ancestorMoved(AncestorEvent event) {}
...
      */
     private void loadTargetComponent() {
-        new Thread() {
-            @Override
-            public void run() {
-                JComponent targetComponent = getTargetComponent();
-
+        new Thread(() -> {
+            JComponent targetComponent = getTargetComponent();
+            SwingUtilities.invokeLater(() -> {
```
ahadas, line 109: can't we drop the new thread from here now that we register the code execution in SwingUtilities#invokeLater? the call to getTargetComponent should be a quick one..

mucommander-viewer-image/build.gradle
```diff
@@ -6,17 +6,35 @@ dependencies {
...
+
+    // image plugins to handle additional types (like psd) or faster alternatives to built-in ones (jpeg)
+    comprise 'com.twelvemonkeys.imageio:imageio-jpeg:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-tiff:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-psd:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-webp:3.11.0'
+    // requirements for plugins
+    comprise 'com.twelvemonkeys.common:common-lang:3.11.0'
+    comprise 'com.twelvemonkeys.common:common-io:3.11.0'
+    comprise 'com.twelvemonkeys.imageio:imageio-metadata:3.11.0'
```
ahadas, line 19: I guess few of these libs are retrieved for the pdf bundle as well, right? I wonder if we can introduce it as a separate bundle(s)

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/Activator.java
```diff
@@ -18,24 +18,44 @@
...
-
-        viewerRegistration = context.registerService(FileViewerService.class, new ImageFileViewerService(), null);
+        try {
+            IIORegistry registry = IIORegistry.getDefaultInstance();
+            registry.registerServiceProvider(new JPEGImageReaderSpi());
+            registry.registerServiceProvider(new PSDImageReaderSpi());
+            registry.registerServiceProvider(new TIFFImageReaderSpi());
+            registry.registerServiceProvider(new WebPImageReaderSpi());
+        } catch (Exception e) {
+            LOGGER.error("Error registering additional image service providers", e);
```
ahadas, line 55: shall we exit in this case?

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/ImageViewer.java
```diff
@@ -86,7 +95,7 @@ class ImageViewer implements FileViewer, ActionListener {
     private JScrollPane scrollPane = new JScrollPane();
     private ImageStatusPanel statusPanel = new ImageStatusPanel();
 
-    private InitialZoom initialZoom = InitialZoom.RESIZE_WINDOW;
+    private InitialZoom initialZoom = InitialZoom.FIT_TO_WINDOW;
```
ahadas, line 98: what's the initial size of the window?

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/ImageViewer.java
```diff
@@ -309,17 +318,33 @@ public void extendMenu(JMenuBar menuBar) {
...
-                image = imageViewerPanel.getToolkit().createImage(bout.toByteArray());
+        long start = System.currentTimeMillis();
+        BufferedImage image;
+
+        try (ByteArrayOutputStream bout = new ByteArrayOutputStream();
+             InputStream in = file.getInputStream()) {
+
+            StreamUtils.copyStream(in, bout);
+            try (ImageInputStream input = new ByteArrayImageInputStream(bout.toByteArray())) {
+                Iterator<ImageReader> readers = ImageIO.getImageReaders(input);
```
ahadas, line 329: can it be that we fail with one reader and succeed with a different one?

ahadas (commented):
please also drop a line in the readme file

mucommander-core/src/main/java/com/mucommander/ui/layout/AsyncPanel.java
pskowronek: I would say that we don't want to block Swing threads - we run in a separate thread as much as we can then any UI related things are being done by Swing thread it feels is OK to update UI ("This method should be used when an application thread needs to update the GUI. ").

mucommander-viewer-image/build.gradle
pskowronek: Yes, a separate bundle grouping Twelve Monkey would be nice. But, I'm not sure if I have enough nerves to do that :)

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/Activator.java
pskowronek: Noe, and I'm not sure if it can throw any exception at this point. This is rather to be on safe-side and continue to work. If some of those providers won't register itself, then still be the built-in ones (in JRE), and since the list of files that can be previewed is taken dynamically the Preview will still work (for some of files).

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/ImageViewer.java
pskowronek: I will check

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/ImageViewer.java
pskowronek: I will check

pskowronek: I think it is default size for JPanel (but not 100% sure)

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/ImageViewer.java
pskowronek: PR updated

mucommander-core/src/main/java/com/mucommander/ui/layout/AsyncPanel.java
ahadas: right, but do we know that creating and running a thread is quicker than calling getTargetComponent? I don't think we really care about msec - so if it's comparable (or in other words, if we don't see a clear benefit of initiating another thread), I'd drop the new thread here..

mucommander-viewer-image/src/main/java/com/mucommander/viewer/image/Activator.java
ahadas: we can raise an exception here but yeah, if the bundle doesn't start we won't have any support of images - so let's keep it this way

mucommander-core/src/main/java/com/mucommander/ui/layout/AsyncPanel.java
pskowronek: > we know that creating and running a thread is quicker than calling getTargetComponent?

For example, https://github.com/mucommander/mucommander/blob/c58c4c7a4d002735227ba92c5681c3e527828f8e/mucommander-core/src/main/java/com/mucommander/ui/viewer/FileFrame.java#L93 - this can be painfully slow operation when opening large file (as text or image).
Spotted that during debugging.

ahadas: okok, you're right - I wrote this comment without looking again at AsyncPanel and I forgot that this is the async part that is supposed to do the heavy lifting (it looked like an innocent method that returns a Swing component)

ahadas (approved):
I'll try to extract twelvemonkeys to a separate bundle as a follow up to this PR, other than that it lgtm

PR #1231, opened 2024-07-08
Organize imports and some generics

olivierpaul:
Generics

### DIFF
--- a/mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
+++ b/mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
@@ -37,7 +37,6 @@
 
 package com.sun.nfs;
 
-import java.io.*;
 import java.util.*;
 import com.sun.rpc.*;
 
--- a/mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
+++ b/mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
@@ -37,6 +37,7 @@
 
 package com.sun.nfs;
 
+import java.io.*;
 import java.util.*;
 import com.sun.rpc.*;
 
--- a/src/main/java/com/mucommander/main/AutoProcessor.java
+++ b/src/main/java/com/mucommander/main/AutoProcessor.java
@@ -22,7 +22,6 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.StringTokenizer;
@@ -96,12 +95,12 @@ public class AutoProcessor
      * @param configMap Map of configuration properties.
      * @param context The system bundle context.
     **/
-    public static void process(Map configMap, BundleContext context)
+    public static void process(Map<String, String> configMap, BundleContext context)
     {
         ExecutorService executor = null;
         try {
             executor = Executors.newFixedThreadPool(8);
-            var configMapLocal = (configMap == null) ? new HashMap() : configMap;
+            var configMapLocal = (configMap == null) ? new HashMap<String, String>() : configMap;
             processAutoDeploy(executor, configMapLocal, context);
             processAutoProperties(executor, configMapLocal, context);
         } finally {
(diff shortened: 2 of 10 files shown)


src/main/java/com/mucommander/main/AutoProcessor.java
```diff
@@ -96,12 +95,12 @@ public class AutoProcessor
...
      * @param context The system bundle context.
     **/
-    public static void process(Map configMap, BundleContext context)
+    public static void process(Map<String, String> configMap, BundleContext context)
     {
         ExecutorService executor = null;
         try {
             executor = Executors.newFixedThreadPool(8);
-            var configMapLocal = (configMap == null) ? new HashMap() : configMap;
+            var configMapLocal = (configMap == null) ? new HashMap<String, String>() : configMap;
```
ahadas, line 103: how about replacing it with something like the following before the try-block:
```yaml
if (configMap == null)
    configMap = new HashMap<>();
```

ahadas (commented):
I originally tried to avoid changing AutoProcessor and rather change the example it was taken from in apache-felix, but I recall @pskowronek found it challenging to make changes there so I'm overall ok with these changes. one comment inside

src/main/java/com/mucommander/main/AutoProcessor.java
pskowronek: No objections. However, since we already modify AutoProcessor maybe let's just set this argument to NotNull (the only usage of this `process` method is from `muCommander.java`, and map is already enforced to be not null: https://github.com/mucommander/mucommander/blob/50749ce332462fd36d35aaea1b9e5edf04e52e65/src/main/java/com/mucommander/main/muCommander.java#L271)

mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
```diff
@@ -37,7 +37,6 @@
 
 package com.sun.nfs;
 
-import java.io.*;
 import java.util.*;
 import com.sun.rpc.*;
```
pskowronek, line 41: what about removing wildcard imports completely?

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas: yep, good idea

mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
ahadas: I would prefer to keep this code as close as possible to the original code that we cloned, I'll revert changes in "cloned" code

pskowronek: OK. What about other places with wildcards?
FYI, I've just picked this file randomly to make a comment. 
BTW, such subsystems/packages could be marked as being cloned from elsewhere? For example in `package-info.java` with a link to a source?

src/main/java/com/mucommander/main/AutoProcessor.java
```diff
@@ -95,18 +97,15 @@ public class AutoProcessor
...
-            executor = Executors.newFixedThreadPool(8);
-            var configMapLocal = (configMap == null) ? new HashMap<String, String>() : configMap;
-            processAutoDeploy(executor, configMapLocal, context);
-            processAutoProperties(executor, configMapLocal, context);
-        } finally {
-            if (executor != null) {
-                executor.shutdown();
-            }
+        var executor = Executors.newFixedThreadPool(8);
+        try (AutoCloseable closable = executor::shutdown){
```
pskowronek, line 103: or simply:
`try (var executor = Executors.newFixedThreadPool(8)) {' 

However, Executors became AutoCloseable in java 19, right?

mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
ahadas: > OK. What about other places with wildcards? FYI, I've just picked this file randomly to make a comment.

ah ok. in general, I'm good with replacing imports with wildcards - the formatter also does it, but as we discussed once - I prefer not to have PRs with so many formatting changes as they are hard to review. PRs that just replace those imports or replacing them gradually would be ok

> BTW, maybe such subsystems/packages could be marked as being cloned from elsewhere? For example in `package-info.java` with a link to a source?

yeah, that sounds like a good idea

src/main/java/com/mucommander/main/AutoProcessor.java
ahadas: right

mucommander-protocol-nfs/src/main/java/com/sun/nfs/NfsSecurity.java
ahadas: > > BTW, maybe such subsystems/packages could be marked as being cloned from elsewhere? For example in `package-info.java` with a link to a source?
> 
> yeah, that sounds like a good idea

(for a separate PR)

Issue #1218, opened 2024-06-17
[Bug] muCommander fails to open folders with a lot of .rar files.


ErikG-GH:
### Is there an existing issue for this?

- [X] I have searched the existing issues (including the closed ones)

### Description

When I try to open folder that contains many rar files (>100, >200?) muCommander shows a spinning blue wheel but does nothing further. Previously I was on version 1.3.0 of muCommander but recently I installed version 1.4.0 and the failing to open started. I just installed 1.4.1 but the problem persists.
What is also annoying is that when I go back to 1.3.0 the problem remains.
MuCommander is slow in opening folders that contain some dozens of rar files but eventually opens the folders and I can also open the rar files. Above a certain amount of rar files muCommander fails to open the folder.

### Steps to reproduce

Just fill a folder with many rar files and try to open it.

### Expected behavior

I expect to be able to open a folder with any amount of rar files and previously I could in version 1.3.0 before I updated.

### Actual behavior

What happens is that the blue wheel keeps spinning. I can move my mouse around and click on another folder or panel but there is no execution of any command. I have to close down muCommander which is perfectly possible. I don't need to force a stop.

### muCommander version

muCommander
Version: 1.4.1
Build date: 2024/06/12
Build number: a65fe3e

### Java version

Java
Runtime version: 23-internal
VM name: OpenJDK 64-Bit Server VM
VM version: 23-internal-ahadas-0081ac197a656cdc0d88d5a5e2a0f399ee023654
VM vendor: Oracle Corporation

### Operating System type and version

OS
Name: Mac OS X
Version: 13.4
Architecture: aarch64

### Relevant log output

```shell
When I try to open a folder and it fails I can see the debug window but not a lot happens there and I can't copy any of it because I can't execute any command in muCommander at that moment.
I don't know where I can access a log file.
```

ErikG-GH:
Another attempt.
I opened muCommander displaying the 'ebooks newbooks/' folder and then tried to open the subfolder '2024/' .
In the Thread-10 nothing happens anymore. I waited a couple of minutes before closing muCommander.
Here a short portion of the log:

2024-06-17 15:28:16.461 [WT-EventQueue-0] DEBUG com.mucommander.core.LocationChanger:163 - folder=/Volumes/EBooks/ebooks/ebooks newbooks/2024/ selectThisFileAfter=null
2024-06-17 15:28:16.464 [Thread-10] DEBUG c.m.core.BrowseLocationThread:155 - starting folder change...
[...]

PR #1237, opened 2024-07-15
Patch sevenzipbindings for M1 Mac

ShayArtzi:
This PR adds M1 support for `sevenzipbindings`. It's a kind of non-traditional short-term fix, I would appreciate any feedback/suggestions for improvement.

## The problem
 `sevenzipbindings` did not have a new release for quite some time. The [last release](https://github.com/borisbrodski/sevenzipjbinding/releases/tag/Release-16.02-2.01) was in the beginning of 2020, and this is what muC is currently using.

In this release, the `sevenzipjbinding-all-platforms` artifact does not contain a native library for M1 Mac (`Mac-aarch64` support).

This impacts muC's ability to open `rar` and `7z` files under M1.

## Temporary solution
Since it does not look like a new release is coming anytime soon (see [this issue](https://github.com/borisbrodski/sevenzipjbinding/issues/63)), I wanted to try and patch the latest release to include `Mac-aarch64` support.

I compiled `sevenzipbindings` on an M1 Mac according to the instructions [here](https://github.com/borisbrodski/sevenzipjbinding/issues/47#issuecomment-2037593450). Then, I patched the original `sevenzipjbinding-all-platforms` to include the compiled artifact. I was then able to confirm 7z and rar files are working as expected on M1. I also got a positive feedback from @Andreas0602 [here](https://github.com/mucommander/mucommander/issues/1218#issuecomment-2227761018).

The next thing I did was to look for a way to easily publish a maven package. I cloned [szb](https://github.com/ShayArtzi/sevenzipjbinding) and was hoping to use github packages to serve the patched jar, but unfortunately it looks like it's not possible to have a public facing package registry that does not require any authentication.

I then created a mirror [gitlab repository](https://gitlab.com/shayartzi/sevenzipjbinding) for the above repo, and was able to [publish the package over there](https://gitlab.com/shayartzi/sevenzipjbinding/-/packages/27273903).

Finally, I changed the dependency in the relevant grade file, and I was able to build the application and confirm the patched file is being pulled from gitlab.

## Long term solution
Once szb will have a new release, we can just go back to use its releases, since the only change in the patched version is the native library addition (no java code changes at all).

### DIFF
--- a/sevenzipjbindings/build.gradle
+++ b/sevenzipjbindings/build.gradle
@@ -1,6 +1,8 @@
 repositories {
     repositories.mavenCentral()
     maven {
+        // To be able to use a patched version of sevenzipjbinding-all-platforms
+        // From the following repo: https://gitlab.com/shayartzi/sevenzipjbinding/-/packages/27273903
         url = uri("https://gitlab.com/api/v4/projects/59936705/packages/maven")
     }
 }
@@ -9,7 +11,9 @@ dependencies {
     api project(':mucommander-commons-file')
 
     comprise 'net.sf.sevenzipjbinding:sevenzipjbinding:16.02-2.01'
-    // comprise 'net.sf.sevenzipjbinding:sevenzipjbinding-all-platforms:16.02-2.01'
+
+    // Use a patched version of sevenzipjbinding-all-platforms for Mac M1 support.
+    // See https://github.com/mucommander/mucommander/pull/1237
     comprise 'com.github.shayartzi.sevenzipjbinding:sevenzipjbinding-all-platforms:16.02-2.01'
 
     testImplementation 'org.testng:testng:7.10.2'
--- a/sevenzipjbindings/build.gradle
+++ b/sevenzipjbindings/build.gradle
@@ -1,7 +1,6 @@
 repositories {
     repositories.mavenCentral()
     maven {
-        // To be able to use a patched version of sevenzipjbinding-all-platforms
         url = uri("https://www.mucommander.com/maven/")
     }
 }
--- a/sevenzipjbindings/build.gradle
+++ b/sevenzipjbindings/build.gradle
@@ -2,8 +2,7 @@ repositories {
     repositories.mavenCentral()
     maven {
         // To be able to use a patched version of sevenzipjbinding-all-platforms
-        // From the following repo: https://gitlab.com/shayartzi/sevenzipjbinding/-/packages/27273903
-        url = uri("https://gitlab.com/api/v4/projects/59936705/packages/maven")
+        url = uri("https://www.mucommander.com/maven/")
     }
 }
 
@@ -14,7 +13,7 @@ dependencies {
 
     // Use a patched version of sevenzipjbinding-all-platforms for Mac M1 support.
     // See https://github.com/mucommander/mucommander/pull/1237
-    comprise 'com.github.shayartzi.sevenzipjbinding:sevenzipjbinding-all-platforms:16.02-2.01'
+    comprise 'com.mucommander:sevenzipjbinding-all-platforms:16.02-2.01'
 
     testImplementation 'org.testng:testng:7.10.2'
 }
--- a/sevenzipjbindings/build.gradle
+++ b/sevenzipjbindings/build.gradle
@@ -1,10 +1,16 @@
-repositories.mavenCentral()
+repositories {
+    repositories.mavenCentral()
+    maven {
+        url = uri("https://gitlab.com/api/v4/projects/59936705/packages/maven")
+    }
+}
 
 dependencies {
     api project(':mucommander-commons-file')
 
     comprise 'net.sf.sevenzipjbinding:sevenzipjbinding:16.02-2.01'
-    comprise 'net.sf.sevenzipjbinding:sevenzipjbinding-all-platforms:16.02-2.01'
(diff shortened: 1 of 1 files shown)


ahadas (commented):
@ShayArtzi thanks for posting this, really nice work on that issue
if all we need is to compile the library on arm, did you consider cloning the clode to the `mucommander` org and produce the artifact with github actions (like we do for the JRE now)?
it's not that great to depend on a jar that was created manually and put in such external repo

ShayArtzi:
Thanks for the feedback! I fully understand the concern and I just wanted to show this possible solution.

Running the full build process is not a simple task, as the all-platform jar contains native libraries compiled in different environments. That's the reason why I think patching is the easiest approach.

Some caveats:
- If we publish the artifact somewhere in github (under muC or any other org), it's not going to be possible for developers to pull the dependency easily for local builds. 
- If we make it a part of the build process, we must run on M1 (which is not ideal... for example in my day to day I use windows)

One solution I can offer is to include the patched jar as a "binary" living with the actual code - will this be acceptable?

ahadas:
> Running the full build process is not a simple task, as the all-platform jar contains native libraries compiled in different environments. That's the reason why I think patching is the easiest approach.

patching is definitely easier but we had such jar that were created "manually" in the past, and it was difficult to recall how to create them after some time - that's why I'd prefer to have something automated, even if it's not a complete build. but ok, let's concentrate on where to place the jar first.

> 
> Some caveats:
> 
> * If we publish the artifact somewhere in github (under muC or any other org), it's not going to be possible for developers to pull the dependency easily for local builds.

why wouldn't it be possible to developers to pull the dependency easily for local builds? I've created a maven repository in https://github.com/ahadas/maven - if you change the URL to `https://github.com/ahadas/maven/raw/main`, it would be possible to pull it easily for local builds

> * If we make it a part of the build process, we must run on M1 (which is not ideal... for example in my day to day I use windows)

yeah, and github has hosted runners of this type that we build the dmg for aarch64 on

> 
> One solution I can offer is to include the patched jar as a "binary" living with the actual code - will this be acceptable?

I think we better avoid adding such binaries
how about adding a maven repository in, e.g., `https://www.mucommander.com/maven`?

ShayArtzi:
> > Running the full build process is not a simple task, as the all-platform jar contains native libraries compiled in different environments. That's the reason why I think patching is the easiest approach.
> 
> patching is definitely easier but we had such jar files that were created "manually" in the past, and it was difficult to recall how to create them after some time - that's why I'd prefer to have something automated, even if it's not a complete build. but ok, let's concentrate on where to place the jar first.
> 
> > Some caveats:
> > 
> > * If we publish the artifact somewhere in github (under muC or any other org), it's not going to be possible for developers to pull the dependency easily for local builds.
> 
> why wouldn't it be possible to developers to pull the dependency easily for local builds? I've created a maven repository in https://github.com/ahadas/maven - if you change the URL to `https://github.com/ahadas/maven/raw/main`, it would be possible to pull it easily for local builds
> 
> > * If we make it a part of the build process, we must run on M1 (which is not ideal... for example in my day to day I use windows)
> 
> yeah, and github has hosted runners of this type that we build the dmg for aarch64 on
> 
> > One solution I can offer is to include the patched jar as a "binary" living with the actual code - will this be acceptable?
> 
> I think we better avoid adding such binaries how about adding a maven repository in, e.g., `https://www.mucommander.com/maven`?

If we just work with a simple file hosting as mvn repo - sure this would work. What I meant to say is that it is not possible with github package feature which is a full fledged mvn repo (including the ability to publish artifacts etc).

I'm open to hosting the patched jar under any location you prefer - would you like me to submit a PR for mucommander.github.io ?

ahadas:
> If we just work with a simple file hosting as mvn repo - sure this would work. What I meant to say is that it is not possible with github package feature which is a full fledged mvn repo (including the ability to publish artifacts etc).

ack

> I'm open to hosting the patched jar under any location you prefer - would you like me to submit a PR for mucommander.github.io ?

yes please

ShayArtzi:
NP - I will work on that a bit later. I really appreciate the feedback and I'm excited to fix that :)

sevenzipjbindings/build.gradle
```diff
@@ -1,10 +1,19 @@
-repositories.mavenCentral()
+repositories {
+    repositories.mavenCentral()
+    maven {
+        // To be able to use a patched version of sevenzipjbinding-all-platforms
```
ahadas, line 4: I don't see much value in this comment as the groupId for sevenzipjbinding-all-platforms below indicates why we need this, what do you think?

sevenzipjbindings/build.gradle
```diff
@@ -1,10 +1,19 @@
...
+    }
+}
 
 dependencies {
     api project(':mucommander-commons-file')
 
     comprise 'net.sf.sevenzipjbinding:sevenzipjbinding:16.02-2.01'
-    comprise 'net.sf.sevenzipjbinding:sevenzipjbinding-all-platforms:16.02-2.01'
+
+    // Use a patched version of sevenzipjbinding-all-platforms for Mac M1 support.
```
ahadas, line 14:
```suggestion
    // Use a patched version of sevenzipjbinding-all-platforms with support for Apple silicon
```

sevenzipjbindings/build.gradle
ShayArtzi: I wanted to make the intent super clear but I can remove it

ahadas (approved):

PR #1241, opened 2024-07-19
Drop 'attempt to load non-existing icon' log for actions without icons

ahadas:
Add NoIcon annotation to explicitly state that no icon is set for a particular action, and in case the annotation is set for an action, we don't create a warning of 'attempt to load non-existing icon' for that action.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/NoIcon.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/NoIcon.java
@@ -1,0 +1,9 @@
+package com.mucommander.ui.action;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface NoIcon {
+
+}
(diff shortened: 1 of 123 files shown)


mucommander-core/src/main/java/com/mucommander/ui/action/NoIcon.java
```diff
@@ -0,0 +1,9 @@
+package com.mucommander.ui.action;
```
pskowronek, line 1: copyright note?

mucommander-core/src/main/java/com/mucommander/ui/action/NoIcon.java
```diff
@@ -0,0 +1,9 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
pskowronek, line 6: a javadoc explaining what is the idea behind?

ahadas:
@pskowronek sorry, I missed your comments but I posted #1242 to address them

pskowronek:
> @pskowronek sorry, I missed your comments but I posted #1242 to address them

np, I added them after the merge. Thanks.

Issue #1252, opened 2024-08-16
[Bug] zip file from linux asks for password


aluedtke7:
### Is there an existing issue for this?

- [X] I have searched the existing issues (including the closed ones)

### Description

A zip file created on a linux machine (OpenMediaVault 7, Debian based), can't be opened on an Apple silicon macBook pro with M2max CPU (latest OS Sonoma 14.6.1). When the zip file is double clicked in mucommander on the mac, the **Password-protected file?** dialog pops up. 
The zip file was created with this command on the linux machine: `zip test test.txt`. 
When the zip file is created like this: `zip -X test test.txt` the resulting zip file can be opened with mucommander on the mac without problems. The `-X` option tells the zip command to **exclude extra file attributes**.
When I click on `abort` in the password dialog, I get an error dialog with these details:
```
java.lang.RuntimeException
 at com.mucommander.ui.event.LocationManager.setCurrentFolder(LocationManager.java:104)
 at com.mucommander.core.LocationChanger.setCurrentFolder(LocationChanger.java:352)
 at com.mucommander.core.BrowseLocationThread.run(BrowseLocationThread.java:383)

```
[test.zip](https://github.com/user-attachments/files/16633790/test.zip)

### Steps to reproduce

Double click the attached zip file

### Expected behavior

Should show the zip content without asking for a password

### Actual behavior

Always asks for a password that doesn't exist

### muCommander version

muCommander
Version: 1.4.1
Build date: 2024/06/12
Build number: a65fe3e

### Java version

Java
Runtime version: 23-internal
VM name: OpenJDK 64-Bit Server VM
VM version: 23-internal-ahadas-0081ac197a656cdc0d88d5a5e2a0f399ee023654
VM vendor: Oracle Corporation

### Operating System type and version

OS
Name: Mac OS X
Version: 14.6.1
Architecture: aarch64

ShayArtzi:
I was able to reproduce the issue locally. Seems to me this is a side effect of #1204. I will further look into that and publish a fix PR soon.

aluedtke7:
Thank you for the super fast response and fix!

ShayArtzi:
Sure thing :)

ahadas:
Version 1.5.1 that includes a fix for this issue [has been released](https://github.com/mucommander/mucommander/releases/tag/1.5.1-1)

PR #1253, opened 2024-08-16
Add null check for zip file extended timestamp entry

ShayArtzi:
This PR addresses #1252.

In a case where extended timestamp entry exists in a zip entry but `FLAG_MODTIME` is not present, the `javaTime` field of `ExtendedTimestampExtraField` is going to be null. This in turn lead to NullPointerException being thrown from `ZipEntry::getTimeExtended`.


ahadas (approved):
thanks @ShayArtzi !

PR #1258, opened 2024-09-04
Added proper icons for macOS Sonoma

danielerez:
Menu Bar:
Menu Bar
Dock:
Dock


ahadas:
@danielerez the icons look good but what about their license and credit to whom we get it from?

danielerez:
> @danielerez the icons look good but what about their license and credit to whom we get it from?

The design was uploaded without a specific license: https://macosicons.com/#/u/imacboy
But seems that crediting the creator is enough? https://github.com/elrumo/macOS_Big_Sur_icons_replacements/discussions/1459
Where it can be noted in the repo?

Issue #1326, opened 2025-02-10
[Bug] Open with... doesn't work for muCommander for Apple Silicon


akluch:
### Is there an existing issue for this?

- [x] I have searched the existing issues (including the closed ones)

### Description

When I right click on any file in muCommander and click Open with... it shows "Enable apps here?"

### Steps to reproduce

Prerequisites:
1. Install duti: _brew install duti_
2. Make sure that PATH contains /opt/homebrew/bin and /opt/homebrew/sbin (since zsh is my default shell on Mac, I specify PATH in .zshrc. Just in case I also specified PATH in .bashrc as well)
3. Verified via muCommander's ToggleTerminal that paths are set indeed.
4. Run duti -h in ToggleTerminal and make sure it displays duti help information.

Steps to reproduce:
1. Launch muCommander
2. Right click on any file in muCommander and click Open with...

The exact same version of muCommander but for Mac with Intel chip - works fine without any issue.

### Expected behavior

It should show the list of apps to associate that particular file type with.

### Actual behavior

It shows "Enable apps here?" despite duti being installed.

### muCommander version

1.5.2

### Java version

Runtime version: 24-ea

### Operating System type and version

Mac OS X
Version: 15.3

### Relevant log output

```shell
[2025-02-09 23:34:54.267] SEVERE OSXDesktopAdapter.java#runCommand,490 Unexpected result from running: '[/bin/zsh, -c, which duti]', timed out?: false, exit code: 1

[2025-02-09 23:34:54.308] SEVERE OSXDesktopAdapter.java#runCommand,490 Unexpected result from running: '[/bin/zsh, -l, -c, which duti]', timed out?: false, exit code: 1

[2025-02-09 23:34:54.308] SEVERE OSXDesktopAdapter.java#getPathOfDutiCmd,464 Command 'duti' not found
```

akluch:
Any updates on this?

Thanks

pskowronek:
> Any updates on this?

Are you sure `duti` is on the PATH? Can you open Terminal and execute `duti` command?

akluch:
> > Any updates on this?
> 
> Are you sure `duti` is on the PATH? Can you open Terminal and execute `duti` command?

Yep. I ran 'duti  -h' from Terminal and got usage info for it.

pskowronek:
OK, when you open Terminal from muC (F12) - are you able to execute `duti`, or any other commands installed via brew. BTW, have installed duti using brew?

[...]

PR #1332, opened 2025-03-28
Use type instead of which while checking duti + logs #1326

pskowronek:
Use `type` instead of `which` while checking `duti` command presence. 
`type` is preferred to check a command (see https://stackoverflow.com/a/56516482)
Additionally, added stderr output in debug mode.

Trying to fix the following issue  #1326


ahadas (approved):
thanks @pskowronek, this change looks good regardless of the referenced issue but yeah, let's see if it shed more light on that issue

PR #1336, opened 2025-04-13
fix identification of 'duti' and add debug log

ahadas:
apparently, 'type' on macOS 15.4 doesn't seem to respect the '-p' flag and returns "duti is /opt/homebrew/bin/duti" for 'type -p -f duti'
    
this leads to the following error: Error executing command: [duti is /opt/homebrew/bin/duti, -h]. Error msg: Cannot run program "duti is /opt/homebrew/bin/duti": error=2, No such file or directory
    
thus, replacing 'type -p -f' with 'command -v'

also add more information in the log that may help investigating #1326


pskowronek (approved):
Great finding!

Issue #1338, opened 2025-04-13
[Bug] Exceptions in logs in the current master


pskowronek:
### Is there an existing issue for this?

- [x] I have searched the existing issues (including the closed ones)

### Description

While running muC built from the current sources, the following exceptions are being logged:

1) On start:
```
>> ERROR: Bundle tomcat-servlet-api [190] Error starting file:/Users/[CUT CUT CUT]/build/osgi/bundle/tomcat-servlet-api-8.5.61.jar (org.osgi.framework.BundleException: Unable to resolve tomcat-servlet-api [190](R 190.0): missing requirement [tomcat-servlet-api [190](R 190.0)] osgi.wiring.package; (osgi.wiring.package=org.apache.catalina.connector) Unresolved requirements: [[tomcat-servlet-api [190](R 190.0)] osgi.wiring.package; (osgi.wiring.package=org.apache.catalina.connector)])
org.osgi.framework.BundleException: Unable to resolve tomcat-servlet-api [190](R 190.0): missing requirement [tomcat-servlet-api [190](R 190.0)] osgi.wiring.package; (osgi.wiring.package=org.apache.catalina.connector) Unresolved requirements: [[tomcat-servlet-api [190](R 190.0)] osgi.wiring.package; (osgi.wiring.package=org.apache.catalina.connector)]
        at org.apache.felix.framework.Felix.resolveBundleRevision(Felix.java:4398)
        at org.apache.felix.framework.Felix.startBundle(Felix.java:2308)
        at org.apache.felix.framework.Felix.setActiveStartLevel(Felix.java:1566)
        at org.apache.felix.framework.FrameworkStartLevelImpl.run(FrameworkStartLevelImpl.java:297)
        at java.base/java.lang.Thread.run(Thread.java:1575)
[...]
ERROR: Bundle tomcat-servlet-api [190] Error starting/stopping bundle. (org.osgi.framework.BundleException: Unable to resolve tomcat-servlet-api [190](R 190.0): missing requirement [tomcat-servlet-api [190](R 190.0)] osgi.wiring.package; (osgi.wiring.package=org.apache.catalina.connector) Unresolved requirements: [[tomcat-servlet-api [190](R 190.0)] osgi.wiring.package; (osgi.wiring.package=org.apache.catalina.connector)])
[...]

PR #1339, opened 2025-04-14
Fix stacktrace while starting muC #1338 and default system arch when buidling DMG

pskowronek:
1) Fix stacktrace while starting muC #1338 (1st stacktrace). I don't have S3 to check if it works without that dependency.
2) default system arch when buidling DMG - can we have system arch as a default? 

@ahadas please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Then, you can run a debugger that connects to this port using your favorite IDE 
 ### Packaging
 The creation of a DMG file for macOS (produced in build/distributions):
 ```
-./gradlew clean dmg -PskipDmgSign
+./gradlew clean dmg -PskipDmgSign -Parch=[x86_64|aarch64]
 ```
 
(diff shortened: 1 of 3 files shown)


ShayArtzi:
@pskowronek, if you want to test the s3 connector locally, I've used  [minio](https://github.com/minio/minio) in the past and I was able to connect to is from muC - it's very easy to spin up

pskowronek:
> @pskowronek, if you want to test the s3 connector locally, I've used [minio](https://github.com/minio/minio) in the past and I was able to connect to is from muC - it's very easy to spin up

Thanks, I will give it a try in a couple of days.

pskowronek:
@ShayArtzi Tested with minio - muC works with it with the changes from this PR

ShayArtzi:
@pskowronek awesome, this PR looks good to me :)

ahadas (approved):

PR #1340, opened 2025-04-14
Fix stacktraces when browsing protected resources #1338

pskowronek:
Fix stacktraces when browsing protected resources #1338 + updated package/readme.txt

### DIFF
--- a/package/readme.txt
+++ b/package/readme.txt
@@ -47,6 +47,9 @@ Localization:
 Bug fixes:
 - The Start Menu group the application is placed in is now named 'muCommander' instead of 'Unknown' on Windows.
 - The application sometimes cannot start after being killed.
+- Fixed unnecessary stack-traces in the logs while starting or running the appliaction.
+- Bumped JediTerm and its dependencies to the newest version
+- Bumped RSyntaxTextArea to the newest version
 
 Known issues:
 - Some translations may not be up-to-date.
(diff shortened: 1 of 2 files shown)


package/readme.txt
```diff
@@ -47,6 +47,9 @@ Localization:
 Bug fixes:
 - The Start Menu group the application is placed in is now named 'muCommander' instead of 'Unknown' on Windows.
 - The application sometimes cannot start after being killed.
+- Fixed unnecessary stack-traces in the logs while starting or running the appliaction.
```
ahadas, line 50: I'd say it's an improvement rather than a fix, wdyt?

package/readme.txt
```diff
@@ -47,6 +47,9 @@ Localization:
...
+- Bumped JediTerm and its dependencies to the newest version
+- Bumped RSyntaxTextArea to the newest version
```
ahadas, line 52: yeah, other projects use to mention bumps of dependencies but we usually don't do that, unless it fixes a bug that is filed on muCommander. @pskowronek is there a particular reason to mention this? or maybe it can also be documented as an improvement

ahadas (commented):
the code changes look good, commented on the readme file

package/readme.txt
pskowronek: I will move those 2 entries to improvements. For RSyntaxTextArea they added some features like support for VHDL and some bugfixes. As it comes to JediTerm - it is hard to tell what they do, but I think they constantly improve integration things (I hope :>).

package/readme.txt
pskowronek: ok, moving to improvements section.

pskowronek:
PR updated

ahadas (approved):

PR #1341, opened 2025-04-16
Checking for 'duti' command now checks well known paths first, and uses -i as last resort #1326

pskowronek:
Checking for 'duti' command now checks well known paths first, and uses -i as last resort #1326


ahadas (approved):
nice!
how about documenting this as an improvement in the readme file?

pskowronek:
> nice! how about documenting this as an improvement in the readme file?

Done!

Issue #1327, opened 2025-02-13
[Bug] Transfer of files hangs when there is an asterisk present in the file name


Dino6702:
### Is there an existing issue for this?

- [x] I have searched the existing issues (including the closed ones)

### Description

The transfer process hangs when it encounters an * in the file name. Can't even skip have to cancel  the operation

### Steps to reproduce

Transfer a file from remote to locally with an * in the filename. The  filename that caused the issue was ALLOC*CA

### Expected behavior

The transfer should complete without error or hanging.

### Actual behavior

The transfer hangs

### muCommander version

Version: 1.5.2
Build date: 2024/10/18
Build number: 4d5c14d

### Java version

Runtime version: 23-internal
VM name: OpenJDK 64-Bit Server VM
VM version: 23-internal-ahadas-22e68b4be09cb27d2b8d6c4df79104b9ce55a14b
VM vendor: Oracle Corporation

### Operating System type and version

Windows 11
10.0
amd64

### Relevant log output

```shell

```

ShayArtzi:
This is probably due to `*` is not a legal character for NTFS file system - to support copying such files over we will how to replace the illegal character with some other character - I will look into some options

PR #1346, opened 2025-04-26
Handle illegal NTFS chars

ShayArtzi:
This is a possible solution for #1327.

When copying over files from a remote location into a windows machine, if there is an NTFS illegal char, muC will fail copying it over. In this PR I added logic to replace these chars with ASCII percent encoding.

### DIFF
--- a/mucommander-commons-file/src/main/java/com/mucommander/commons/file/AbstractFile.java
+++ b/mucommander-commons-file/src/main/java/com/mucommander/commons/file/AbstractFile.java
@@ -30,6 +30,9 @@ import java.security.NoSuchAlgorithmException;
 
 import javax.swing.Icon;
 
+import com.mucommander.commons.file.protocol.FileProtocols;
+import com.mucommander.commons.file.util.WindowsFileNameSanitizer;
+import com.mucommander.commons.runtime.OsFamily;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -799,6 +802,11 @@ public abstract class AbstractFile implements FileAttributes {
      */
     public AbstractFile getChild(String relativePath, AbstractFile template) throws IOException {
         FileURL childURL = (FileURL)getURL().clone();
+
+        if (childURL.getScheme().equals(FileProtocols.FILE) && OsFamily.WINDOWS.isCurrent()) {
+            relativePath = WindowsFileNameSanitizer.sanitizeFileName(relativePath);
+        }
+
         childURL.setPath(addTrailingSeparator(childURL.getPath())+ relativePath);
 
         return FileFactory.getFile(childURL, true);
--- a/mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
+++ b/mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
@@ -1,0 +1,32 @@
+package com.mucommander.commons.file.util;
+
+import java.util.Set;
+
+public class WindowsFileNameSanitizer {
+
+    private static final Set<Character> ILLEGAL_CHARS = Set.of(
+            '<', '>', ':', '"', '/', '\\', '|', '?', '*', '`'
+    );
+
+    public static String sanitizeFileName(String input) {
+        StringBuilder sanitized = new StringBuilder();
+
+        for (char c : input.toCharArray()) {
+            if (ILLEGAL_CHARS.contains(c)) {
+                sanitized.append(String.format("%%%02X", (int) c));
+            } else {
+                sanitized.append(c);
+            }
+        }
+
+        return sanitized.toString();
+    }
+
+    // Example usage
+    public static void main(String[] args) {
+        String original = "report<final>*version?.txt";
+        String safe = sanitizeFileName(original);
+        System.out.println("Sanitized: " + safe);
+    }
+
+}
--- a/mucommander-commons-file/src/test/java/com/mucommander/commons/file/util/WindowsFileNameSanitizerTest.java
+++ b/mucommander-commons-file/src/test/java/com/mucommander/commons/file/util/WindowsFileNameSanitizerTest.java
@@ -1,0 +1,16 @@
+package com.mucommander.commons.file.util;
+
+import org.junit.Assert;
+import org.testng.annotations.Test;
+
+public class WindowsFileNameSanitizerTest {
+
+    @Test
+    public void sanitizeFileName_whenCalledWithIllegalChars_thenSanitizeCorrectly() {
+        String input = "report<final>*version?.txt";
+        String expected = "report%3Cfinal%3E%2Aversion%3F.txt";
+
+        Assert.assertEquals(expected, WindowsFileNameSanitizer.sanitizeFileName(input));
+    }
+
+}


mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
```diff
@@ -0,0 +1,32 @@
+package com.mucommander.commons.file.util;
```
pskowronek, line 1: copyright

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
```diff
@@ -0,0 +1,32 @@
...
+public class WindowsFileNameSanitizer {
+
+    private static final Set<Character> ILLEGAL_CHARS = Set.of(
+            '<', '>', ':', '"', '/', '\\', '|', '?', '*', '`'
+    );
+
+    public static String sanitizeFileName(String input) {
+        StringBuilder sanitized = new StringBuilder();
+
+        for (char c : input.toCharArray()) {
```
pskowronek, line 14: use Streams?

pskowronek:
Have you checked how other file managers behave? Like Midnight Commander, Double Commander? Do they also replace characters in the same way?

pskowronek:
Hmm, not sure if silently renaming things is ideal. Ideally, it would be nice to present a dialog box (similar to File Collision) to give the user options: skip, rename/sanitize (plus a choice either with numerical representation as you have right now or with  underscores), plus checkbox 'Apply to all'. However, as a short term fix the proposed change is fine especially that fixes greater problem with hanging transfers. Additionally I would add logs or .warn of .info level that sanitization was performed.

mucommander-commons-file/src/main/java/com/mucommander/commons/file/AbstractFile.java
```diff
@@ -799,6 +802,11 @@ public final AbstractFile getChild(String relativePath) throws IOException {
      */
     public AbstractFile getChild(String relativePath, AbstractFile template) throws IOException {
         FileURL childURL = (FileURL)getURL().clone();
+
+        if (childURL.getScheme().equals(FileProtocols.FILE) && OsFamily.WINDOWS.isCurrent()) {
```
ahadas, line 806: we already have few Windows-specific things in the LocalFile class, how about moving this logic there?

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
```diff
@@ -0,0 +1,32 @@
+package com.mucommander.commons.file.util;
+
+import java.util.Set;
+
+public class WindowsFileNameSanitizer {
```
ahadas, line 5:
```suggestion
public class WindowsFilenameSanitizer {
```

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final Set<Character> ILLEGAL_CHARS = Set.of(
+            '<', '>', ':', '"', '/', '\\', '|', '?', '*', '`'
+    );
+
+    public static String sanitizeFileName(String input) {
+        StringBuilder sanitized = new StringBuilder();
+
+        for (char c : input.toCharArray()) {
+            if (ILLEGAL_CHARS.contains(c)) {
+                sanitized.append(String.format("%%%02X", (int) c));
```
ahadas, line 16: I see that as @pskowronek commented before, it's more common to replace invalid characters with underscore - I think it's a better option for us

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
```diff
@@ -0,0 +1,32 @@
...
+            } else {
+                sanitized.append(c);
+            }
+        }
+
+        return sanitized.toString();
```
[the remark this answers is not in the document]
ahadas: yeah, chatgpt also suggests additional changes:
```Java
public static String sanitizeFilename(String filename) {
    // Replace illegal characters with an underscore
    String sanitized = filename.replaceAll("[\\\\/:*?\"<>|]", "_");

    // Trim spaces at the start and end
    sanitized = sanitized.trim();

    // Prevent reserved names (like CON, PRN, etc.) — optional safety
    String[] reservedNames = {
        "CON", "PRN", "AUX", "NUL",
        "COM1", "COM2", "COM3", "COM4", "COM5", "COM6", "COM7", "COM8", "COM9",
        "LPT1", "LPT2", "LPT3", "LPT4", "LPT5", "LPT6", "LPT7", "LPT8", "LPT9"
    };
    for (String reserved : reservedNames) {
        if (sanitized.equalsIgnoreCase(reserved)) {
            sanitized = "_" + sanitized; // Or append an underscore
            break;
        }
    }

    // Windows max filename length is 255 characters
    if (sanitized.length() > 255) {
        sanitized = sanitized.substring(0, 255);
    }

    return sanitized;
}
```

I'm ok with merging the posted changes (including the test) to fix the reported bug, but it would be nice to to extend this class and add few more tests. @ShayArtzi for your consideration

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    // Example usage
+    public static void main(String[] args) {
+        String original = "report<final>*version?.txt";
+        String safe = sanitizeFileName(original);
+        System.out.println("Sanitized: " + safe);
+    }
```
ahadas, line 30: I'd suggest to drop this - it's quite obvious and we have unit test(s)..

ahadas (commented):
@ShayArtzi looks good overall as I see there's no built-in or a good third party function we can use for this

ShayArtzi:
> Hmm, not sure if silently renaming things is ideal. Ideally, it would be nice to present a dialog box (similar to File Collision) to give the user options: skip, rename/sanitize (plus a choice either with numerical representation as you have right now or with underscores), plus checkbox 'Apply to all'. However, as a short term fix the proposed change is fine especially that fixes greater problem with hanging transfers. Additionally I would add logs or .warn of .info level that sanitization was performed.

I agree - let me add some logging. I was thinking about this as a quickfix and not necessarily a long-term solution. Keep in mind it's kind of a rare edge case - you have to combine windows and a non-windows remote AND have one of these unique characters in file names (which is usually a bad practice anyways... I would never name a file with an `*`, that's basically asking for troubles)

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
ShayArtzi: yep that's a leftover... let me take it out

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
ShayArtzi: Underscore is what total commander is doing as well, and if you both think it's a better approach, I'm open to changing it. 

The case for using ASCII percent encoding is: (a) we don't lose information and (b) reduce chance for file name collision on copy.

Let me know what you think - the change itself is trivial of course

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
ShayArtzi: I can do something like:
```java
public static String sanitizeFileName(String input) {
    return input.chars()
        .mapToObj(c -> ILLEGAL_CHARS.contains((char) c) 
            ? String.format("%%%02X", c) 
            : String.valueOf((char) c))
        .collect(Collectors.joining());
}
```

That's probably a bit less efficient but in practice should work fine. LMK what you think - I'm open to using either

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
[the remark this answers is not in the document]
ShayArtzi: Yeah let's start with my changes and then handle more cases as needed (and also consider an interactive rename approach like @pskowronek is suggesting)

mucommander-commons-file/src/main/java/com/mucommander/commons/file/AbstractFile.java
ShayArtzi: let me do some homework and I'll report back :)

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
ShayArtzi: I might be missing something but what is the change here?

mucommander-commons-file/src/main/java/com/mucommander/commons/file/AbstractFile.java
```diff
@@ -799,6 +802,16 @@ public final AbstractFile getChild(String relativePath) throws IOException {
...
+            String sanitizedRelativePath = WindowsFileNameSanitizer.sanitizeFileName(relativePath);
+            if (!relativePath.equals(sanitizedRelativePath)) {
+                LOGGER.warn("Windows file renamed [relativePath = {}, sanitizedRelativePath = {}]",
```
ShayArtzi, line 809: @pskowronek I've added a WARN here

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
ahadas: long time ago I asked someone that reported issues to muCommander to stop asking us to change certain things just because other file managers do them that way, but even though I see the benefits you've mentioned, imho in this case it makes sense to align with what others do

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
ShayArtzi: got you - changed

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/WindowsFileNameSanitizer.java
ShayArtzi: you got it - changed to `_`

ShayArtzi:
> @ShayArtzi sorry again for the time it took me to review this I've relocated the code that is specific to local-files and optimized the imports in AbstractFile (after the former change, the only changes to AbstractFile are related to imports, but that's fine) could you please test that I didn't break something in the process? please also post an update to the readme file in a separate PR

Sure, I'll post an update later this week.

ShayArtzi:
I confirmed this is still working by copying a file from an SFTP remote into my local machine. Here you can see the `*` is replaced with a `_`:
hkFdSndz4J

Once this PR is merged I will publish another PR for the readme update.

PR #1351, opened 2025-05-12
Binary viewer library update

hajdam:
Update of bined library
- might slightly improve rendering performance on Windows


ahadas (approved):
thanks @hajdam

PR #1352, opened 2025-05-28
Fix Path Traversal Vulnerability in Recursive File Deletion

th555555:
This PR addresses a critical security vulnerability in the `deleteRecursively` method that allows attackers to delete arbitrary files on the file system through path traversal attacks.

### Before (Vulnerable)
- No boundary validation for file deletion
- Missing protection against path traversal attacks
- Vulnerable to symlink attacks and ".." directory traversal

### After (Secure)
- **Base directory validation** - requires base directory parameter
- **Canonical path checking** - resolves symlinks and relative paths
- **Path boundary enforcement** - prevents deletion outside base directory
- **Enhanced error handling** - clear security violation messages

### DIFF
--- a/mucommander-commons-file/src/main/java/com/mucommander/commons/file/AbstractFile.java
+++ b/mucommander-commons-file/src/main/java/com/mucommander/commons/file/AbstractFile.java
@@ -1331,13 +1331,29 @@ public abstract class AbstractFile implements FileAttributes {
      * or not implemented by the underlying filesystem.
      */
     protected final void deleteRecursively(AbstractFile file) throws IOException, UnsupportedFileOperationException {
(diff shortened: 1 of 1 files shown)


ahadas (commented):
@th555555 there're few issues with this PR:
1. we probably don't want to raise an exception when finding a file that is out of the tree of the removed folder but just not to delete its children
2. there shouldn't be a case where the "baseDir" is unknown (null) - it's supposed to be the directory that deleteRecursively was originally called for
3. we cannot rely on Java `Path` in this context, that's an AbstractFile and most of its implementations don't return a Java `Path`

but more broadly, I don't understand why there's a security vulnerability here since we don't list symlinks - what's the risk in terms of path traversal you see with the current code?

Issue #1372, opened 2025-10-26
[Bug] macOS - delayed switching between Firefox and muC


pskowronek:
### Is there an existing issue for this?

- [x] I have searched the existing issues (including the closed ones)

### Description

While quickly switching between muCommander and other apps using CMD-Tab, there's a delay before muC shows up again.

### Steps to reproduce

- open Firefox and muCommander
- switch to muCommander
- cmd-tab to Firefox and immediately once again cmd-tab to muC

### Expected behavior

muCommander shows up back again immediatelly

### Actual behavior

There's delay before muCommander shows up. 

### muCommander version

- 1.5.2
- current master

### Java version

muC built-in

### Operating System type and version

- macOS 15.6.1 (24G90) (apple silicon)

### Relevant log output

```shell

```

PR #1373, opened 2025-10-26
Fix for switching lag between Firefox (GPU app) and muC #1372

pskowronek:
Fix for switching lag between Firefox (i.e. GPU-using app, like web-renderers -> Thunderbird or Chrome) and muC #1372.

It seems that macOS metal and JDK still have some glitches. Disabling `sun.java2d.metal` apparently fixes the problem.


ahadas:
ah I see in the issue that it's `muC built-in` - interestingly, it doesn't seems to happen on my laptop

PR #1375, opened 2025-11-17
Fix undesired beep and undesired character when closing Terminal with F12, fix similar to #1187

pskowronek:
- Fix undesired beep and undesired character when closing Terminal with F12 - similar to #1187
- Bumped JediTerm to the newest available version


ahadas (commented):
thanks @pskowronek!

PR #1384, opened 2025-12-19
In Gnome environment, to detect the mouse click interval : if the gco…

olivierpaul:
…nftool command is not present, use gsettings command


ahadas (approved):
I've added some cleanup in separate commits

ahadas:
@olivierpaul thanks!

Issue #1397, opened 2026-01-13
[Bug] Sorting by name does not sort alphabetically


rhstanton:
### Is there an existing issue for this?

- [x] I have searched the existing issues (including the closed ones)

### Description

Create directories 20_test, 60_thematic, 80_thematic, and 201_test, and 2012_13. When I sort these by name, I want the directories beginning with 6 and 8 to appear after all the directories with names beginning with 2. But they don't! I see

20_test
60_thematic
80_thematic
201_test

Similarly, 202_test appears before 2013_14.

It looks like the directories are being sorted in *numerical* order based on the numbers at the start of the names (though I haven't done enough testing to be 100% sure). Plain alphabetic sorting seems like it would give more expected results.

Thanks!

### Steps to reproduce

See above.

### Expected behavior

See above.

### Actual behavior

See above.

### Screenshots?

<img width="951" height="207" alt="Image" src="https://github.com/user-attachments/assets/1de93fa0-3dfd-4535-a116-a00b4cf4825d" />

### muCommander version

1.6.0

### Java version

Runtime version: 24-ea
VM name: OpenJDK 64-Bit Server VM
VM version: 24-ea+13-1421
VM vendor: Oracle Corporation

### Operating System type and version

macOS 26.2

### Relevant log output

```shell

```

ShayArtzi:
Hi @rhstanton!

Looking at the code, I wouldn't say that's a bug but an intentional behavior (see [ref](https://github.com/mucommander/mucommander/blob/eeada78a6328d4b9c7f30ea5d0594ff2fab18ba6/mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/FileComparator.java#L161)).

We can consider adding a setting/knob to control this behavior if that would be useful...

rhstanton:
That would be great. Thanks!Sent from my mobile deviceOn Jan 13, 2026, at 16:02, Shay Artzi ***@***.***> wrote:﻿ShayArtzi left a comment (mucommander/mucommander#1397)
Hi @rhstanton!
Looking at the code, I wouldn't say that's a bug but an intentional behavior (see ref).
We can consider adding a setting/knob to control this behavior if that would be useful...

PR #1398, opened 2026-01-14
Introduce custom sort order

ShayArtzi:
This PR addresses #1397 - adding a setting to control the file name sort logic. I would appreciate any feedback/suggestions/thoughts.

Here is what I did:
1. Fixed unit tests for `mucommander-commons-file` - tests were not actually running (Gradle was using Junit and not TestNG). Some of the tests were broken - I was able to fix some and ignore others
2. Added LEXICOGRAPHIC and NATURAL sort support + unit tests
3. UI change - adding an setting

image

Here is a demo of the feature:
muc_order

### DIFF
--- a/mucommander-preferences/src/main/java/com/mucommander/conf/MuPreferences.java
+++ b/mucommander-preferences/src/main/java/com/mucommander/conf/MuPreferences.java
@@ -94,9 +94,9 @@ public class MuPreferences implements MuPreferencesAPI {
     public static final String  SET_DROP_ACTION_TO_COPY           = "change_drop_action_to_copy";
     /** Default file drag and drop behavior */
     public static final boolean DEFAULT_SET_DROP_ACTION_TO_COPY   =  true;
-
+    /** Whether to sort files lexicographically (or naturally) */
     public static final String FILES_LEXICOGRAPHIC_ORDER = "lexicographic_order_of_files";
-
+    /** Default value for {@link #FILES_LEXICOGRAPHIC_ORDER} */
     public static final boolean DEFAULT_FILES_LEXICOGRAPHIC_ORDER = false;
 
     // - Log variables -------------------------------------------------------
--- a/mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTableModel.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/main/table/FileTableModel.java
@@ -707,7 +707,7 @@ public class FileTableModel extends AbstractTableModel {
                 sortInfo.getFoldersFirst(),
                 getNameFunc(),
                 locale,
-                fileComparatorMode );
+                fileComparatorMode);
     }
 
 
--- a/package/readme.txt
+++ b/package/readme.txt
@@ -36,6 +36,7 @@ What's new since v1.5.2 ?
 
 New features:
 - SMBv2 and SMBv3 are now supported.
+- Introduced lexicographical sorting for filenames.
 
 Improvements:
 - When selecting a different shell, the change takes effect immediately in the terminal.


mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/FileComparatorModeEnum.java
```diff
@@ -0,0 +1,12 @@
+package com.mucommander.commons.file.util;
```
ahadas: please add the copyright at the top

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/FileComparator.java
```diff
@@ -54,15 +54,17 @@
...
-    private boolean directoriesFirst;
+    private final boolean directoriesFirst;
     /** Returns the value for the 'name' column for a file */
-    private Function<AbstractFile, String> nameFunc;
+    private final Function<AbstractFile, String> nameFunc;
     /** Locale that is used to sort by filenames */
-    private Locale locale;
+    private final Locale locale;
+
+    private final FileComparatorModeEnum mode;
```
ahadas: please document this property

mucommander-preferences/src/main/java/com/mucommander/conf/MuPreferences.java
```diff
@@ -95,6 +95,10 @@ public class MuPreferences implements MuPreferencesAPI {
     /** Default file drag and drop behavior */
     public static final boolean DEFAULT_SET_DROP_ACTION_TO_COPY   =  true;
 
+    public static final String FILE_COMPARATOR_USE_LEXICOGRAPHIC_SORT = "file_comparator_use_lexicographic_sort";
+
+    public static final boolean DEFAULT_FILE_COMPARATOR_USE_LEXICOGRAPHIC_SORT = false;
```
ahadas: that's interesting - I have a TODO item that I can't find time to tackle which is to introduce per-OS keyboard shortcuts, I think we'd need to differentiate some configuration items like this one as well if there are operating systems on which lexicographic order is the default. but for now setting its default to `false` makes sense

ShayArtzi:
> thanks @ShayArtzi, good exploration of what exists and what can be done to address this! few comments inside

Thanks for the review!

1. Let me address the comments related to the new feature
2. In the meantime I would like to discuss the tests I've ignored:

When I discovered unit tests were not running, I couldn't really figure out why some them failed. In some cases the tested logic just didn't make sense to me so it's hard to fix. I didn't want to go into the rabbit hole of fixing these since I figured the tests did not run for quite some time and the original behavior probably changed over time. I also figured it best to have the majority of the tests restored and ignore some vs ignoring them all...

Would you like to take a look at the original tests and LMK what you think? by removing the override methods the original test should run again and fail.

BTW:
1. There were cases where I did fix broken tests since it made sense to me, like ignoring a POSIX related exception on Windows.
2. The reason why I was looking into the tests was because I wanted to write my own test for the sort logic ;)

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/FileComparatorModeEnum.java
ShayArtzi: Moved the file

mucommander-commons-file/src/test/java/com/mucommander/commons/file/protocol/local/LocalFileURLTest.java
```diff
@@ -173,4 +173,10 @@ protected String getPathSeparator() {
     protected boolean isQueryParsed() {
         return false;
     }
+
+    // @Test
+    @Override
+    public void testEquals() throws MalformedURLException {
+
```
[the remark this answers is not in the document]
ShayArtzi: same as before :)

mucommander-commons-file/src/main/java/com/mucommander/commons/file/util/FileComparator.java
ShayArtzi: done

mucommander-preferences/src/main/java/com/mucommander/conf/MuPreferences.java
```diff
@@ -95,6 +95,10 @@ public class MuPreferences implements MuPreferencesAPI {
     /** Default file drag and drop behavior */
     public static final boolean DEFAULT_SET_DROP_ACTION_TO_COPY   =  true;
 
+    public static final String FILE_COMPARATOR_USE_LEXICOGRAPHIC_SORT = "file_comparator_use_lexicographic_sort";
```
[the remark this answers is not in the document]
ShayArtzi: done

mucommander-core/src/main/java/com/mucommander/ui/dialog/pref/general/MiscPanel.java
```diff
@@ -193,6 +195,12 @@ public boolean hasChanged() {
         setDropActionToCopyCheckBox.addDialogListener(parent);
         northPanel.add(setDropActionToCopyCheckBox);
 
+        fileComparatorModeCheckBox = new PrefCheckBox(Translator.get("prefs_dialog.file_comparator_use_lexicographic_sort"), () -> MuConfigurations.getPreferences().getVariable(
+                MuPreference.FILE_COMPARATOR_USE_LEXICOGRAPHIC_SORT,
+                MuPreferences.DEFAULT_FILE_COMPARATOR_USE_LEXICOGRAPHIC_SORT));
+        fileComparatorModeCheckBox.addDialogListener(parent);
```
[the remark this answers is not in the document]
ShayArtzi: Let's go for a description change for now and improve the UI in the future - is that okay with you?

If so, could you please suggest the best wording to use?

Issue #643, opened 2022-05-06
File panel stops being usable after Drag-n-drop under macOS


pskowronek:
File panel stops being usable after Drag-n-drop under macOS

To reproduce:
- open a directory in the panel
- mark file or files and drag-n-drop them to Finder (or photos.google.com in Chrome)
- return to muCommander, try to move the selection on this panel with cursors on keyboard - the selection doesn't want to move (it returns the selected file)
- the other panel works fine
- to fix the problem, one must try to drag-n-drop a file but without finalizing the drop - this way muCommander can know that the operation of drag-n-drop is finished (this is what I imagine) and selection can be moved no problem.

The root cause may or may not be the same as for #43 

muCommander version: 1.0.0-1, macOS: 10.15.7, java: 17.0.1+12 (OpenJDK 64-Bit Server VM)

ahadas:
hi @pskowronek, I'm not able to reproduce it - the default drag-and-drop operation is now 'move', do the dragged files disappear from the panel after a short time? it would be best to record a video that demonstrates this process - maybe I do something differently

pskowronek:
But the issue is about drag-n-drop into other app, like Firefox or Chrome, it's not about drag-n-drop between panels. I will try to record a video later on.

pskowronek:
I've tried for 20 minutes and nothing! Let's put this on hold until I experience that once again :/

ahadas:
ack, so I'm closing and feel free to reopen if it happens again

pskowronek:
Still happens on 1.5.2, macOS: 26.2 (25C56). 

This is how it looks like after drag-n-drop of jpg files to Firefox:

![Image](https://github.com/user-attachments/assets/eb5b4b8b-5886-41e4-8d7e-4d160339005d)

On the video above I was using up arrow to move the selector up, and nothing - it doesn't go further than 4 rows. The same with Down key. Going into subfolder, moving selector works over there, when navigating back to parent folder it is still the same.

PR #1402, opened 2026-01-31
Fix for unstable panel after drag and drop

pskowronek:
Fix for unstable panel after Drag and drop #643

The main fix is in FileDragSourceListener#dragDropEnd - after I was unable to reproduce the problem so far.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTargetListener.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTargetListener.java
@@ -45,7 +45,7 @@ import com.mucommander.ui.dialog.file.ProgressDialog;
 import com.mucommander.ui.main.FolderPanel;
 import com.mucommander.ui.main.MainFrame;
 
-import javax.swing.*;
+import javax.swing.SwingUtilities;
 
 /**
  * Provides file(s) 'drop' support to components that add a <code>DropTarget</code> using this


mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDragSourceListener.java
```diff
@@ -177,5 +182,14 @@ public void dragDropEnd(DragSourceDropEvent event) {
         DnDContext.setDragInitiatedByMucommander(false);
         DnDContext.setDragInitiator(null);
         DnDContext.setDragGestureModifiersEx(0);
+
+        // Fix for https://github.com/mucommander/mucommander/issues/643
```
ShayArtzi: Do you think this is specific for macOS? if so, perhaps we should have this `repaint` conditional on OS type?

ahadas: yeah, it's not supposed to be an expensive operation and ending drag operation doesn't happen that frequently so I think it's fine to keep it for all platforms

mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTargetListener.java
```diff
@@ -45,6 +45,8 @@
 import com.mucommander.ui.main.FolderPanel;
 import com.mucommander.ui.main.MainFrame;
 
+import javax.swing.*;
```
ahadas: please avoid this kind of imports

package/readme.txt
```diff
@@ -59,6 +59,7 @@ Bug fixes:
 - When copying files from a non-Windows source, replace characters that are invalid in Windows filenames (e.g., '*', ':') with an underscore.
 - Copying names, basenames, and paths of files to the clipboard stopped working (regression).
 - Fixed undesired beep and character when leaving Terminal with F12.
+- Fixed unstable file panel after Drag and drop.
```
ahadas:
```suggestion
- Fixed unstable file panel after dragging a file and dropping it on a separate application.
```

I think it is also better to replace `unstable` with a more accurate description but I'm not sure I get the full picture of what actually happened in that case

pskowronek:
@ahadas - approve & merge?

mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTargetListener.java
ahadas: ah we already have a commit that changes the source listener here, but nothing changes this for the target listener

ahadas (approved):

Issue #1417, opened 2026-02-19
[Bug] Report a Bug - escaping issues


pskowronek:
### Is there an existing issue for this?

- [x] I have searched the existing issues (including the closed ones)

### Description

Report a Bug from muC menu opens Github issue/bug reporting with improperly escaped characters.

### Steps to reproduce

- In muC go to Help menu, and hit Report a Bug
- Firefox opens with Github bug reporting

### Expected behavior

- Pre-filled data should be properly presented

### Actual behavior

- Pre-filled data are not properly presented

### Screenshots?

<img width="346" height="144" alt="Image" src="https://github.com/user-attachments/assets/268346a3-aeef-42e7-8a22-2aee03fc356e" />

<img width="360" height="445" alt="Image" src="https://github.com/user-attachments/assets/e6678b79-df96-4c13-aa04-d46fb9fbaf5d" />

### muCommander version

muC 1.5.2, 1.6.0 nightly

### Java version

embedded

### Operating System type and version

Mac OS  X 26.3 aarch64

### Relevant log output

```shell

```

PR #1418, opened 2026-02-19
Fix Report a Bug escape problem on macos

pskowronek:
Fix Report a Bug escape problem on macos #1417

Checked only on macOS 26.3. It would be nice to verify the fix on windows and linux :/ (since the urls are opened via LocalProcess / ProcessRunner)

Since I changed CRLF to LF it is hard to see the change in `OpenURLInBrowserAction`:
`InformationDialog.showErrorDialogIfNeeded(getMainFrame().getJFrame(), DesktopManager.browse(new URL((String)url)));` ->
`InformationDialog.showErrorDialogIfNeeded(getMainFrame().getJFrame(), DesktopManager.browse((String)url));`

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/core/desktop/InternalBrowse.java
+++ b/mucommander-core/src/main/java/com/mucommander/core/desktop/InternalBrowse.java
@@ -30,7 +30,7 @@ import com.mucommander.desktop.UrlOperation;
 /**
  * @author Nicolas Rinaudo
  */
-class InternalBrowse extends UrlOperation {
+public class InternalBrowse extends UrlOperation {
     // - Instance variables --------------------------------------------------
     // -----------------------------------------------------------------------
     /** Underlying desktop instance. */
--- a/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ReportBugAction.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/action/impl/ReportBugAction.java
@@ -17,17 +17,21 @@
 
 package com.mucommander.ui.action.impl;
 
-import java.util.Arrays;
-import java.util.Map;
-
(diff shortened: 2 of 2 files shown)


pskowronek:
> on Ubuntu the title still looks good ("[Bug] ") but Java version and the OS version miss the line separator: 21.0.10OpenJDK 64-Bit Server VM21.0.10 7-Ubuntu-125.10Ubuntu Linux6.17.0-14-genericaarch64
> 
> I'll check on Windows as well

Hm, do you know what was the reason of having `BROWSE` declared two times, both with the same priority `SYSTEM_OPERATIONS` in DesktopManager:

https://github.com/mucommander/mucommander/blob/875edc05df468e70188ec7f8ebc44c0f8059a2a5/mucommander-core/src/main/java/com/mucommander/core/desktop/DesktopManager.java#L189-L192

If I comment out the second one, then Report a Bug works properly with the original `URLEncoder.encode` because it uses java built-in mechanism `Desktop.browse()` via `InternalBrowse`. 

My guess is that `CommandBrowse` by executing `open` command may have problems how to properly either tokenize parameters and/or escape params which go through LocalProcess / ProcessRunner which might behave differently on different operating systems?

ahadas:
> If it fixes the problem for macOS (and report bug worked well on Windows and Linux) then I OK with it. What I would add in your fix is the comment why it was added this way (refer the bug on github?). Since I added this feature (like hm 3 years ago) there were some subtle changes in macOS or JRE that required a bit different encoding - I suspect that it was related how macOS `open` worked or sth? Nevertheless, I hope that your fix will be finally bullet proof across macOS/JRE changes and it is more elegant than my workaround.

yeah, I'll add that comment and document the change before merging

PR #1459, opened 2026-04-06
Bumped VAqua to 13 to fix loading problem on macOS

pskowronek:
Bumped VAqua to 13 to fix loading problem on macOS Tahoe

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,7 +69,7 @@ dependencies {
     comprise 'org.apache.felix:org.apache.felix.main:7.0.5'
     comprise 'com.beust:jcommander:1.82'
     comprise 'com.formdev:flatlaf:2.6'
-    comprise 'org.violetlib:vaqua:10'
+    comprise 'org.violetlib:vaqua:13'
 
     compileOnly 'com.google.code.findbugs:jsr305:1.3.9'
 
--- a/mucommander-core/build.gradle
+++ b/mucommander-core/build.gradle
@@ -26,7 +26,7 @@ dependencies {
 
 
     compileOnly 'com.formdev:flatlaf:2.2'
-    compileOnly 'org.violetlib:vaqua:10'
+    compileOnly 'org.violetlib:vaqua:13'
 
     // JediTerm deps: if you change any version below, remember to change it also in jetbrains-jediterm
     compileOnly 'org.jetbrains.jediterm:jediterm-core:3.57'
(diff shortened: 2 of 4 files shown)


ahadas (commented):
thanks @pskowronek!
the changes look good but when testing them I noticed that when switching back from vaqua to another look-and-feel, weird things happen - specifically, the inactive file table appears empty.. 
the current situation where we get an exception on newer macOS versions is not great either but if we see such issues, maybe it's better to drop it temporarily until we see it's ok?

pskowronek:
> thanks @pskowronek! the changes look good but when testing them I noticed that when switching back from vaqua to another look-and-feel, weird things happen - specifically, the inactive file table appears empty.. the current situation where we get an exception on newer macOS versions is not great either but if we see such issues, maybe it's better to drop it temporarily until we see it's ok?

You are right - initially VAqua works ok, but after restart there's problem with table - here only 3 items on the first panel is shown, the second one is OK.  We can keep this PR open for further investigation.

In another PR I've added a bit nicer error handling for LaF: https://github.com/mucommander/mucommander/pull/1460

pskowronek:
@ahadas PoC after suggestion from VAqua author how to handle DnD - might return to it when VAqua releases version with a fix for concurrent modification.

ahadas:
@pskowronek what's the status of this PR? I tested it locally and it seems to work well but since it's been a while since talking about this (sorry about that btw), I'm not sure if we decided to hold this until a new version of vaqua is released

pskowronek:
> @pskowronek what's the status of this PR? I tested it locally and it seems to work well but since it's been a while since talking about this (sorry about that btw), I'm not sure if we decided to hold this until a new version of vaqua is released

AFAIR with this PR we need to wait for a newer release of VAqua :/

ahadas:
> > @pskowronek what's the status of this PR? I tested it locally and it seems to work well but since it's been a while since talking about this (sorry about that btw), I'm not sure if we decided to hold this until a new version of vaqua is released
> 
> AFAIR with this PR we need to wait for a newer release of VAqua :/

ah for addressing the java.util.ConcurrentModificationException, I see
so how about posting just the part that replaces DropTargetListener with a TransferHandler separately?

pskowronek:
OK, I will create a separate PR for that.

cherry picked to: https://github.com/mucommander/mucommander/pull/1489

PR #1471, opened 2026-04-22
Granting the application permission to access the module 'sun.awt.X11'

douglasimcabral:
To supress the Warning: 'Unexpected error setting WM_CLASS: Unable to make field private static java.lang.String sun.awt.X11.XToolkit.awtAppClassName accessible: module java.desktop does not "opens sun.awt.X11" to unnamed module @35aa28cd'.

This warning occurs because modern Java versions (Java 11 and later) restrict access to internal packages like sun.awt.X11 to improve security.


ahadas (approved):
thanks for posting this @douglasimcabral !

PR #1478, opened 2026-05-17
Move from OSGi to JPMS

ahadas:
Few years ago we started using OSGi for improving modularity, and more importantly to provide a pluggable design for file formats and protocols. We haven't completed the pluggable design and OSGi introduced quite a few technical challenges, and in the meantime the Java Platform Module System became a standard that provides most of the capabilities we need - the only area that OSGi may be better is in loading and unloading bundles at runtime, which iis not that important for us. Thus, we are going to switch from OSGi to JPMS - this transition would likely require few follow ups to fix regressions but we need to start somewhere, and the sooner the better.


ahadas:
@pskowronek @ShayArtzi have you had a chance to look at this? It’s still a work in progress, but I believe it’s the right direction. I'm wondering if we should merge this foundational transition for the upcoming release now, even though there are known gaps (like packaging). merging early would allow us to implement other dependent changes on top of it. wdyt?

pskowronek:
> @pskowronek @ShayArtzi have you had a chance to look at this? It’s still a work in progress, but I believe it’s the right direction. I'm wondering if we should merge this foundational transition for the upcoming release now, even though there are known gaps (like packaging). merging early would allow us to implement other dependent changes on top of it. wdyt?

Yup, great initiative! I was kinda worried that we will stuck with bnd/osgi (to be honest, I've never been a big fan of it :P). 
This is a very good direction. I would merge this as soon as possible so we could understand and feel the mechanics of that new approach.

ShayArtzi:
> @pskowronek @ShayArtzi have you had a chance to look at this? It’s still a work in progress, but I believe it’s the right direction. I'm wondering if we should merge this foundational transition for the upcoming release now, even though there are known gaps (like packaging). merging early would allow us to implement other dependent changes on top of it. wdyt?

I didn't try to run it myself but I agree with @pskowronek - let's go for it, it must be better than OSGI :)

ShayArtzi:
Very exciting!

PR #1479, opened 2026-05-20
Fix terminal font not updating when changed in theme settings

danielerez:
JediTerm caches the font at initialization while theme colors are read lazily on each repaint. Listen for terminal font and color theme changes and reinitialize the active terminal panel so preference updates apply without hiding and reopening the terminal.


mucommander-core/src/main/java/com/mucommander/ui/terminal/TerminalPanelHelper.java
```diff
@@ -0,0 +1,59 @@
...
+final class TerminalPanelHelper {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(TerminalPanelHelper.class);
+
+    private static final Method REINIT_FONT_AND_RESIZE_METHOD;
+
+    static {
+        Method method = null;
+        try {
+            method = TerminalPanel.class.getDeclaredMethod("reinitFontAndResize");
```
ahadas: I don't like this part that uses reflection - as the `reinitFontAndResize` is package-protected, we can try to expose it without reflection

ahadas (approved):

ahadas:
@pskowronek wdyt?

pskowronek:
> @pskowronek wdyt?

LGTM

PR #1485, opened 2026-06-12
zip64 support

### LINKED ISSUE #1484: [Bug] Zip64 file shows up as empty
### Is there an existing issue for this?

- [x] I have searched the existing issues (including the closed ones)

### Description

I have done a google takeout of my data. The data downloads as multiple zip64 files with sizes up to 50 GB. They show up as empty in mucommander, but I can verify via other tools that they contain my data.
I was not(yet) able to produce a testcase without disclosing my own data. (and I am not sure if it's a good idea to upload a 50 gig testcase)

the smaller files (2.9GB and below) from that takeout work.

### Steps to reproduce

(perform a google takeout of your data, in 50GB chunks)
navigate to one of the files in muCommander, press enter, nothing there.

### Expected behavior

zip file contents should be shown.

### Actual behavior

no files show up.

### Screenshots?

_No response_

### muCommander version

Version information:

muCommander
Version: 1.6.2
Build date: 2026/05/14
Build number: 1
Build metadata: 3bc2118

### Java version

Java
Runtime version: 25.0.2
VM name: OpenJDK 64-Bit Server VM
VM version: 25.0.2+10-LTS
VM vendor: Eclipse Adoptium

### Operating System type and version

OS
Name: Linux
Version: 6.8.0-124-generic
Architecture: amd64

### Relevant log output

```shell
(no log output)
```

ShayArtzi:
This address #1484

## Disclaimer 
After debugging the issue myself by hand, I've used claude code to do the actual implementation.

## Background
We basically forked Apache Ant's zip package (which I already tweaked in the past to support extended date/time field). The version of this package we had did not have zip64 support and failed silently. The actual part which failed is the `ZipFile::positionAtCentralDirectory` which simply did not work for zip64. This is because the central directory is found in the very end of the file, and with zip64 this lookup process is a bit different. [Here](https://blog.yaakov.online/zip64-go-big-or-go-home/) is a blog post explaining the issue in a clear way.

## The fix
As mentioned above, I've used claude code to port over the Apache Ant zip64 support to our project. At some point in the future we might consider using a proper library but for the time being and as a quick fix I thought that's a good path to take, especially since AI is doing most of the hard work.

### DIFF
--- a/mucommander-format-zip/src/main/java/com/mucommander/commons/file/archive/zip/provider/CentralDirectoryParsingZipExtraField.java
+++ b/mucommander-format-zip/src/main/java/com/mucommander/commons/file/archive/zip/provider/CentralDirectoryParsingZipExtraField.java
@@ -1,0 +1,46 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package com.mucommander.commons.file.archive.zip.provider;
+
+import java.util.zip.ZipException;
+
+/**
+ * {@link ZipExtraField ZipExtraField} that knows how to parse central directory data.
+ *
+ * <p>--------------------------------------------------------------------------------------------------------------<br>
+ * <br>
+ * This class is based off the <code>org.apache.tools.zip</code> package of the <i>Apache Ant</i> project. The Ant
+ * code has been modified under the terms of the Apache License which you can find in the bundled muCommander license
+ * file.</p>
+ *
+ * @author Apache Ant, Maxence Bernard
+ */
+public interface CentralDirectoryParsingZipExtraField extends ZipExtraField {
+
+    /**
+     * Populate data from this array as if it was in central directory data.
+     * @param data an array of bytes
+     * @param offset the start offset
+     * @param length the number of bytes in the array from offset
+     *
+     * @throws ZipException on error
+     */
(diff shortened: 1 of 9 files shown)


mucommander-format-zip/src/main/java/com/mucommander/commons/file/archive/zip/provider/CentralDirectoryParsingZipExtraField.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * {@link ZipExtraField ZipExtraField} that knows how to parse central directory data.
+ *
+ * <p>--------------------------------------------------------------------------------------------------------------<br>
+ * <br>
+ * This class is based off the <code>org.apache.tools.zip</code> package of the <i>Apache Ant</i> project. The Ant
+ * code has been modified under the terms of the Apache License which you can find in the bundled muCommander license
+ * file.</p>
+ *
+ * @author Apache Ant, Maxence Bernard
```
ahadas: I'm pretty sure Maxence wasn't an author of new files that are added here :)

ahadas (commented):
I'm not sure what's the best way to test it - we don't have enough tests for file formats..
@ShayArtzi did you test this code with enough zip32 archives to have confident on not having regressions by this?
and please add an entry to the readme.txt file for this

mucommander-format-zip/src/main/java/com/mucommander/commons/file/archive/zip/provider/CentralDirectoryParsingZipExtraField.java
ShayArtzi: Good catch. At least I was upfront about using AI ;)

ShayArtzi:
> I'm not sure what's the best way to test it - we don't have enough tests for file formats.. @ShayArtzi did you test this code with enough zip32 archives to have confident on not having regressions by this? and please add an entry to the readme.txt file for this

I understand the concern.

To introduce this fix, I've first created a zip64 file on my machine by packing a directory with more than 4gb of files into a zip archive using [7-zip](https://www.7-zip.org/). I then opened the archive with muC and got an empty directory listing, as expected. I also verified that standard (32bit) zip file works as expected.

I then started my debugging and found the root cause. After Claude backported the Ant zip64 support code, I tried opening both files again and this time both worked. For each file I was able to get the listing and extract the contents.

I thought about adding an automated test to extract a zip64 file but it was not very practical since it will consume a lot of disk space. Since the source is actually cloned from Ant based on the original base64 support change set (backport), I feel okay about this change. Having said that, I think it would be more optimal to migrate away from using our own Ant fork and use a properly maintained library (such as [Apache Commons Compress](https://commons.apache.org/proper/commons-compress/)). This is something I would like to pursue in the future (when I have some time), so I see this PR as a short term fix

ahadas (approved):

PR #1495, opened 2026-07-30
Move args.txt creation to execution phase to fix clean run

ahadas:
The runJpms task was creating args.txt during configuration phase, causing `./gradlew clean run` to fail when clean deleted the file before execution. Moving file creation to doFirst ensures it happens after clean completes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JPMS task execution by generating runtime argument files immediately before execution.
  * Preserved JVM argument handling, including whitespace, quote, and backslash escaping.


ShayArtzi (approved):
LGTM

PR #1501, opened 2026-08-02
Focus on file after pasting a full path into the location bar

### LINKED ISSUE #1498: [Enhancement] Support pasting local file paths into the path field
### Description

I'm a happy long-time MuCommander user, and I'd like to thank you for creating such a great file manager. However, there's one small feature I've been missing for quite a while.

If I have the full path to a local file in my operating system clipboard (for example, copied from IntelliJ IDEA) and paste it into the path field, I would expect MuCommander to navigate to the containing directory and automatically select that file. This would let me immediately perform any further file operations.

Instead, MuCommander currently treats the pasted text as if it were a URL and opens the download dialog. Personally, this behavior has never been useful to me, and it feels unintuitive when pasting a local file path.

Would it be possible to detect local file paths and, when one is pasted into the path field, navigate to the corresponding directory and automatically select the referenced file? I think this would make it much more convenient to work with paths copied from IDEs, terminals, and other applications.

Thank you for considering this suggestion, and thank you for your continued work on MuCommander.

ShayArtzi:
This addresses #1498. Note that ctrl+v didn't actually work on windows so fixing that as well

## Summary by CodeRabbit

* **New Features**
  * Entering or pasting a complete file path in the location bar now navigates to its parent directory and selects the file.
  * Clipboard actions in the location bar remain available alongside file drag-and-drop support.

* **Bug Fixes**
  * Improved handling of regular files by avoiding unnecessary download prompts when their parent directory can be opened.

* **Documentation**
  * Updated the v1.6.3 changelog with the enhanced file-path navigation behavior.

### DIFF
--- a/mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTransferHandler.java
+++ b/mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTransferHandler.java
@@ -17,9 +17,11 @@
 
 package com.mucommander.ui.dnd;
 
+import java.awt.datatransfer.Clipboard;
 import java.awt.datatransfer.DataFlavor;
 import java.awt.event.InputEvent;
 
+import javax.swing.JComponent;
 import javax.swing.TransferHandler;
 
 import org.slf4j.Logger;
@@ -75,19 +77,43 @@ public class FileDropTransferHandler extends TransferHandler {
     private final boolean changeFolderOnlyMode;
 
     /**
+     * The component's previous TransferHandler (if any), e.g. the LAF-installed one that implements
+     * ordinary text cut/copy/paste for a text field. {@code setTransferHandler} replaces the handler
+     * wholesale, and TransferHandler is also what keyboard-triggered clipboard operations (Ctrl+C/X/V)
+     * delegate to - so without forwarding non-drop transfers here, installing this handler on a text
+     * field would silently break its normal clipboard shortcuts.
+     */
+    private final TransferHandler fallbackHandler;
+
+    /**
      * @param folderPanel          the panel to navigate / copy-move into
      * @param changeFolderOnlyMode if {@code true} only folder navigation is performed on drop;
      *                             if {@code false} copy/move is also possible
      */
     public FileDropTransferHandler(FolderPanel folderPanel, boolean changeFolderOnlyMode) {
+        this(folderPanel, changeFolderOnlyMode, null);
+    }
+
+    /**
+     * @param folderPanel          the panel to navigate / copy-move into
+     * @param changeFolderOnlyMode if {@code true} only folder navigation is performed on drop;
+     *                             if {@code false} copy/move is also possible
+     * @param fallbackHandler      the component's previous TransferHandler, to which non-drop transfers
+     *                             (keyboard cut/copy/paste) are forwarded; {@code null} if the component
+     *                             has no clipboard behavior worth preserving (e.g. a plain button)
+     */
+    public FileDropTransferHandler(FolderPanel folderPanel, boolean changeFolderOnlyMode, TransferHandler fallbackHandler) {
         this.folderPanel = folderPanel;
         this.changeFolderOnlyMode = changeFolderOnlyMode;
+        this.fallbackHandler = fallbackHandler;
     }
 
     @Override
     public boolean canImport(TransferSupport support) {
         if (!support.isDrop()) {
-            return false;
+            // Not a drag-and-drop transfer, so this is a keyboard/menu-triggered clipboard paste -
+            // defer to whatever the component normally does with pasted text.
+            return fallbackHandler != null && fallbackHandler.canImport(support);
         }
 
         // Accept any of the three flavors we know how to handle
@@ -111,6 +137,10 @@ public class FileDropTransferHandler extends TransferHandler {
 
     @Override
     public boolean importData(TransferSupport support) {
+        if (!support.isDrop()) {
+            return fallbackHandler != null && fallbackHandler.importData(support);
+        }
+
         if (!canImport(support)) {
             return false;
         }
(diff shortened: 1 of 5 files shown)


mucommander-core/src/main/java/com/mucommander/ui/main/LocationTextField.java
```diff
@@ -281,6 +282,19 @@ public boolean textFieldValidated() {
...
+        // If the entered/pasted location points to an existing non-browsable file (i.e. not a directory
+        // and not an archive that can be browsed into), navigate to its parent folder and select it
+        // there - mirroring how dropping a file onto the location bar is handled (see
+        // FileDropTransferHandler) - instead of falling through to BrowseLocationThread's generic file
+        // handling, which offers to "download" it. Browsable files (archives) keep going through that
+        // generic path so the existing "download or browse" prompt still applies to them.
+        AbstractFile file = FileFactory.getFile(location);
+        if (file != null && file.exists() && !file.isBrowsable()) {
+            AbstractFile parent = file.getParent();
+            if (parent != null)
```
pskowronek: I would encourage to always use braces for all IFs even for single line statements (good practice).
And additionally I would almost always add `else` with `LOGGER.trace` for example (it makes easier to debug or trace execution paths).

pskowronek (commented):
see comment

mucommander-core/src/main/java/com/mucommander/ui/main/LocationTextField.java
ShayArtzi: Good catch, I'm doing some more changes and will address this in my next commit

mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTransferHandler.java
```diff
@@ -149,6 +179,22 @@ public boolean importData(TransferSupport support) {
         return true;
     }
 
+    @Override
+    public int getSourceActions(JComponent c) {
+        // Keyboard-triggered cut/copy (Ctrl+C/X) check this before calling exportToClipboard - forward
+        // it so text selection can still be copied out of a component using this handler.
+        return fallbackHandler != null ? fallbackHandler.getSourceActions(c) : super.getSourceActions(c);
```
ahadas, line 186: do I read this correct as: if fallbackHandler is set, use it since we assume the transfer handler that it was overridden with doesn't provide source action?
I just wonder if we add here logic that is specific to the location text field (and then maybe we should inherit FileDropTransferHandler)...

mucommander-core/src/main/java/com/mucommander/ui/main/LocationTextField.java
```diff
@@ -281,7 +281,9 @@ public boolean textFieldValidated() {
         // Remember that the folder change was initiated by the location field
         folderChangeInitiatedByLocationField = true;
 
-        // Change folder
+        // Change folder. If the entered/pasted location turns out to point to a non-browsable file
+        // rather than a directory, BrowseLocationThread navigates to its parent folder and selects it
+        // there instead of showing the "download" prompt - see the resolution logic in its run() method.
```
ahadas: I think this comment is redundant (it was generated by AI, right?) since it's far from the relevant code and it also refers to the code that existed before (so it won't be so clear for someone that would read it later) - what do you think about dropping it?

ahadas:
looks good overall, two comments inline and please don't forget to document in readme.txt

mucommander-core/src/main/java/com/mucommander/ui/main/LocationTextField.java
ShayArtzi: guilty as charged - that was Claude... sure, I'm removing it :)

mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTransferHandler.java
ShayArtzi: Here is the way I was thinking about it: it's a generic answer to a generic problem - if you replace a component's TransferHandler to add drop support, you silently break whatever that component's TransferHandler used to do for clipboard ops. That's true for any component you attach this to, not just the location bar.

ahadas: yeah, but what I was wondering about is whether we should return `fallbackHandler.getSourceActions(c)` when `fallbackHandler` is set or rather the union of both `fallbackHandler.getSourceActions(c)` and `super.getSourceActions(c)` in the general case

package/readme.txt
```diff
@@ -40,6 +40,7 @@ New features:
 Improvements:
 - Terminal: Font and color updates now apply instantly.
 - Support zip64 files (zip files larger than 4gb).
+- Pasting or typing a full file path into the location bar now navigates to its parent folder and selects the file.
```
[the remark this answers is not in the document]
ahadas: I don't like the use of "regular file" either, I think it should be something like "non-browsable file"

mucommander-core/src/main/java/com/mucommander/ui/dnd/FileDropTransferHandler.java
ShayArtzi: This seems like a bigger change (since I will be changing the return type of the method) and I'm not sure in what case we would need it. IMO, if we ever need it for some reason, we can make the change.

ahadas: @ShayArtzi well, I'm not suggesting to change it now - my point was that if we think this addition won't necessarily be a common case for more widgets then maybe making it a FileDropTransferHandler that is specific to the location text field would make more sense

ShayArtzi: Good point, and it's a fair call either way. I'd lean toward leaving it in the base class for now given how cheap the fallback plumbing is, but happy to revisit if a second use case ever makes the split pay for itself

ahadas (approved):
thanks @ShayArtzi !

